PR #1, opened 2015-01-06
Add support for passing arbitrary flags to test executables.

kjellander-unity:
This also removes the support for specifying multiple binaries,
which shouldn't be too much of an issue since it's easy to script
a loop if multiple binaries are to be executed.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Copyright 2013 Google Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


kjellander-unity:
@pbos can you have a look?

pbos:
I submitted 363e11aa0d70b88b0ef1364687bd10171425a3f9 instead. Does this work for you? Additional arguments must be specified after a '--'.

kjellander-unity:
It doesn't help with the problems we're seeing in http://build.chromium.org/p/client.webrtc.fyi/waterfall unfortunately, since it's passing --no-sandbox without any -- terminator.
Then we'd have to change runtest.py instead...

pbos:
(Make that bc6168b08623a4be83b31e825944bcdca1818d81, had a bug.)

pbos:
Considering this fixed since we found a way to add -- before --no-sandbox gets added. Closing. :)

PR #2, opened 2015-03-31
Speed up tests by running the biggest ones first

kwiberg:
And default to one thread per CPU.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -12,11 +12,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import cPickle
+import gzip
 import optparse
+import os
 import subprocess
 import sys
 import threading
 import time
+import zlib
 
 stdout_lock = threading.Lock()
 class FilterFormat:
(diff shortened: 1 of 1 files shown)


kwiberg:
Updated the pull request. (Patch 2/3: Now store just the latest time for each test in a single file. Patch 3/3: Eliminated the named function.)

gtest-parallel
```diff
@@ -82,6 +87,45 @@ class RawFormat:
   def end(self):
     pass
 
+# Record of test runtimes. Has built-in locking.
+class TestTimes(object):
+  def __init__(self, save_file):
+    "Create new object seeded with saved test times from the given file."
+    self.__old_times = {}  # (test binary, test name) -> runtime in seconds
```
pbos: Do you need both old_times and new_times? get_test_time is only called before tests are running, right?

Also, imo remove the __ prefixes, unless you can point me to somewhere saying they're Google style.

gtest-parallel
```diff
@@ -82,6 +87,45 @@ class RawFormat:
...
+    self.__new_times = {}  # (test binary, test name) -> runtime in seconds
+
+    # Protects __new_times; __old_times doesn't need locking since
+    # it's read-only after construction.
+    self.__lock = threading.Lock()
+
+    try:
+      with gzip.GzipFile(save_file, "rb") as f:
+        self.__old_times = cPickle.load(f)
```
pbos: Is there any chance that you can validate that old_times is now a (test binary, test name) -> runtime in seconds? Or at least (string, string) -> float so we don't get any runtime errors if we change file format.

gtest-parallel
kwiberg: I think you're right regarding old_times and new_times. I'll combine them.

Prefixing with __ isn't about style, it's how member variables are made private in Python. See e.g. https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references.

gtest-parallel
kwiberg: That's a good idea. Will do.

gtest-parallel
```diff
@@ -175,11 +236,12 @@ def run_job((command, job_id, test)):
     logger.log(str(job_id) + '> ' + line.rstrip())
 
   code = sub.wait()
-  runtime_ms = int(1000 * (time.time() - begin))
-  logger.log(str(job_id) + ': EXIT ' + str(code) + ' ' + str(runtime_ms))
+  runtime = time.time() - begin
```
pbos: runtime_ms so we can be consistent, any sub-ms precision doesn't matter either way

PR #3, opened 2015-04-24
Fix issues related to TEST_P.

stefanholmer:
Fixes problems with parameterized test fixtures.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -128,9 +128,11 @@ for test_binary in binaries:
   if options.gtest_also_run_disabled_tests:
     command += ['--gtest_also_run_disabled_tests']
 
-  test_list = subprocess.Popen(command +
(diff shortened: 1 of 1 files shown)


gtest-parallel
```diff
@@ -145,6 +147,9 @@ for test_binary in binaries:
     line = line.strip()
     if not options.gtest_also_run_disabled_tests and 'DISABLED' in line:
       continue
+    line = line.split('#')[0].strip()
+    if not line:
```
pbos, line 151: Did you need this "if not line"?

stefanholmer, line 151: I don't think so, but I think it makes sense to skip a line if it was a comment only, right?

pbos, line 151: whatever, sgtm :)

PR #5, opened 2015-05-19
When a test fails, run it first the next time

kwiberg:
This patch makes run_job() return a very large "elapsed time" for jobs
that fail. The effect is that on the next run, tests that previously
failed will be tried first regardless of their actual runtime; since
tests that failed before are likely to fail again, and since observing
failures is often the reason one runs tests in the first place, this
will usually greatly reduce the time one has to wait.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -221,7 +221,10 @@ logger.log(str(-1) + ': TESTCNT ' + ' ' + str(len(tests)))
 
 exit_code = 0
 
-# Run the specified job. Returns the elapsed time in milliseconds.
+# Run the specified job. Return the elapsed time in milliseconds if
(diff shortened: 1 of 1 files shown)


gtest-parallel
```diff
@@ -221,7 +221,9 @@ logger.log(str(-1) + ': TESTCNT ' + ' ' + str(len(tests)))
 
 exit_code = 0
 
-# Run the specified job. Returns the elapsed time in milliseconds.
+# Run the specified job. Return the elapsed time in milliseconds if
+# the job succeeds, or a very large number (larger than any reasonable
+# elapsed time) if the job fails.
```
pbos: \+ "to make them run first next time."

pbos:
lgtm, squash

PR #6, opened 2015-05-20
Handle Ctrl-C gracefully

kwiberg:
Basically, there are two things we need to do: exiting _gracefully_
without emitting backtraces, and exiting _promptly_ without waiting
for tests currently in progress to finish (or even worse, tests that
haven't even started yet).

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -18,12 +18,58 @@ import gzip
 import multiprocessing
 import optparse
 import os
+import signal
 import subprocess
 import sys
 import tempfile
+import thread
 import threading
 import time
 import zlib
+
+# An object that catches SIGINT sent to the Python process and notices
+# if processes passed to wait() die by SIGINT (we need to look for
+# both of those cases, because pressing Ctrl+C can result in either
+# the main process or one of the subprocesses getting the signal).
+#
+# Before a SIGINT is seen, wait(p) will simply call p.wait() and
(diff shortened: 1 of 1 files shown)


kwiberg:
This is a totally rewritten version that works properly without being dreadfully ugly.

gtest-parallel
```diff
@@ -18,13 +18,62 @@ import gzip
...
+    with self.__lock:
+      self.__on_sigint()
+  def got_sigint(self):
+    with self.__lock:
+      return self.__got_sigint
+  def wait_for_or_terminate(self, p):
+    with self.__lock:
+      if self.__got_sigint:
+        p.terminate()
+        raise self.ProcessWasInterrupted
```
pbos: Remove this line (and wait for the zombie), we'll still raise this on line 73.

pbos: Maybe add a comment instead of this line though, waiting on the terminated process should be fine.

gtest-parallel
```diff
@@ -18,13 +18,62 @@ import gzip
...
+        self.__processes.pop().terminate()
+      except OSError:
+        pass
+  def __sigint_handler(self, signal_num, frame):
+    with self.__lock:
+      self.__on_sigint()
+  def got_sigint(self):
+    with self.__lock:
+      return self.__got_sigint
+  def wait_for_or_terminate(self, p):
```
pbos: Rename wait(self, p)

PR #9, opened 2016-04-10
Fix support for TYPED_TEST_CASE.

bbannier:
Prior to this fix a `TYPED_TEST_CASE` would include the trailing type
comment leading to incorrectly constructed test names.


bbannier:
@googlebot I signed it!

pbos:
Oops. Thanks! Turns out we only have one TYPED_TEST_CASE in our codebase, so we've had one unittest that silently succeeds but doesn't run.

PR #10, opened 2016-04-21
Made sure status updates fit on screen.

bbannier:
When test finish `gtest-parallel`'s reporting makes no attempt to deal with test names which do not fit on the line, but instead overflows. We could try to be a little smarter and truncate the names of the reported tests if needed.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -128,6 +128,16 @@ class FilterFormat:
   failures = []
 
   def print_test_status(self, last_finished_test, time_ms):
+    # Determine usable line width.
+    screen_width = sys.maxsize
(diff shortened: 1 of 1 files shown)


pbos:
@kwiberg can you take a look? I thought we already did this?

kwiberg:
Yes, there's the term_width function on line 79, called bu Outputter on line 100.
@bbannier, what does your new code do that the existing code doesn't?

bbannier:
Sorry for the noise, this is indeed already being taken care of. I was driven to look into this by the test times currently being cut off when the length of the filter exceeds the line length, but the proper way to fix is likely to instead reorder the displayed information (e.g., number, time, filter).

PR #12, opened 2016-11-17
Add support for swarming.

ehlemur-zz:
Add support for [swarming](https://github.com/luci/luci-py) so that the tests can be split between different machines.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -265,12 +265,12 @@ parser.add_option('--gtest_also_run_disabled_tests', action='store_true',
 parser.add_option('--format', type='string', default='filter',
                   help='output format (raw,filter)')
 parser.add_option('--print_test_times', action='store_true', default=False,
-                  help='List the run time of each test at the end of execution')
+                  help='list the run time of each test at the end of execution')
 parser.add_option('--shard_count', type='int', default=1,
-                  help='Total number of shards (for sharding test execution '
+                  help='total number of shards (for sharding test execution '
                        'between multiple machines)')
 parser.add_option('--shard_index', type='int', default=0,
-                  help='Zero-indexed number identifying this shard (for '
+                  help='zero-indexed number identifying this shard (for '
                        'sharding test execution between multiple machines)')
 
 (options, binaries) = parser.parse_args()
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -338,6 +338,9 @@ for test_binary in binaries:
     tests.append((times.get_test_time(test_binary, test),
                   test_binary, test, command))
 
+# GTEST_SHARD_INDEX and GTEST_TOTAL_SHARDS must be removed from the environment
+# otherwise it will be picked up by the binary, causing a bug where only tests
+# in the firsh shard are executed.
 test_env = os.environ.copy()
 test_env.pop('GTEST_SHARD_INDEX', None)
 test_env.pop('GTEST_TOTAL_SHARDS', None)
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -266,6 +266,15 @@ parser.add_option('--format', type='string', default='filter',
                   help='output format (raw,filter)')
 parser.add_option('--print_test_times', action='store_true', default=False,
                   help='When done, list the run time of each test')
+parser.add_option('--shard_count', type='int',
+                  default=int(os.environ.pop('GTEST_TOTAL_SHARDS', 1)),
+                  help=('Total number of shards (for sharding test execution '
+                        'between multiple machines). Default: %default'))
+parser.add_option('--shard_index', type='int',
+                  default=int(os.environ.pop('GTEST_SHARD_INDEX', 0)),
+                  help=('Zero-indexed number identifying this shard (for '
+                        'sharding test execution between multiple machines). '
+                        'Default: %default'))
 
 (options, binaries) = parser.parse_args()
 
@@ -280,6 +289,14 @@ elif options.format == 'filter':
   logger = FilterFormat()
 else:
   sys.exit("Unknown output format: " + options.format)
+
+if options.shard_count < 1:
+  sys.exit("Invalid number of shards: %d. Must be at least 1." %
+           options.shard_count)
+if options.shard_index < 0 or options.shard_count <= options.shard_index:
+  sys.exit("Invalid shard index: %d. Must be between 0 and %d." %
+           (options.shard_index, options.shard_count - 1))
+
 
 # Find tests.
 save_file = os.path.join(os.path.expanduser("~"), ".gtest-parallel-times")
@@ -320,6 +337,8 @@ for test_binary in binaries:
       continue
     tests.append((times.get_test_time(test_binary, test),
                   test_binary, test, command))
+
+tests = tests[options.shard_index::options.shard_count]
 
 if options.failed:
   # The first element of each entry is the runtime of the most recent
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -294,7 +294,8 @@ if options.shard_count < 1:
   parser.error("Invalid number of shards: %d. Must be at least 1." %
                options.shard_count)
 if not (0 <= options.shard_index < options.shard_count):
-  parser.error("Invalid shard index: %d. Must be between 0 and %d." %
+  parser.error("Invalid shard index: %d. Must be between 0 and %d "
+               "(less than the number of shards)." %
                (options.shard_index, options.shard_count - 1))
 
 # Find tests.
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -267,11 +267,11 @@ parser.add_option('--format', type='string', default='filter',
 parser.add_option('--print_test_times', action='store_true', default=False,
                   help='When done, list the run time of each test')
 parser.add_option('--shard_count', type='int',
-                  default=int(os.environ.pop('GTEST_TOTAL_SHARDS', 1)),
+                  default=int(os.environ.get('GTEST_TOTAL_SHARDS', 1)),
                   help=('Total number of shards (for sharding test execution '
                         'between multiple machines). Default: %default'))
 parser.add_option('--shard_index', type='int',
-                  default=int(os.environ.pop('GTEST_SHARD_INDEX', 0)),
+                  default=int(os.environ.get('GTEST_SHARD_INDEX', 0)),
                   help=('Zero-indexed number identifying this shard (for '
                         'sharding test execution between multiple machines). '
                         'Default: %default'))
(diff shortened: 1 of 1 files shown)


gtest-parallel
```diff
@@ -266,6 +266,15 @@ parser.add_option('--format', type='string', default='filter',
                   help='output format (raw,filter)')
 parser.add_option('--print_test_times', action='store_true', default=False,
                   help='When done, list the run time of each test')
+parser.add_option('--shard_count', type='int',
+                  default=int(os.environ.pop('GTEST_TOTAL_SHARDS', 1)),
```
kwiberg: Why not os.environ.get?

gtest-parallel
```diff
@@ -266,6 +266,15 @@ parser.add_option('--format', type='string', default='filter',
...
+                  help=('Total number of shards (for sharding test execution '
+                        'between multiple machines). Default: %default'))
+parser.add_option('--shard_index', type='int',
+                  default=int(os.environ.pop('GTEST_SHARD_INDEX', 0)),
+                  help=('Zero-indexed number identifying this shard (for '
+                        'sharding test execution between multiple machines). '
+                        'Default: %default'))
```
kwiberg: I'm not really fond of defaulting to environment variables. Users can't discover it without reading the source, and what's worse, when you call gtest-parallel in a script there's no indication that it does magical things depending on the values of some environment variables.

Would it be terribly inconvenient to require callers that want to use sharding to simply specify these options instead?

gtest-parallel
```diff
@@ -281,6 +290,14 @@ elif options.format == 'filter':
 else:
   sys.exit("Unknown output format: " + options.format)
 
+if options.shard_count < 1:
+  sys.exit("Invalid number of shards: %d. Must be at least 1." %
+           options.shard_count)
+if options.shard_index < 0 or options.shard_count <= options.shard_index:
```
kwiberg: You only have to test options.shard_count <= options.shard_index, since you already know that options.shard_count >= 1.

kwiberg (changes requested):
Nice. Some comments though.

gtest-parallel
ehlemur-zz: It would introduce a bug, where only the tests in the first shard are executed.
We need to pop these environment variables before executing the tests.

gtest-parallel executes each test into it's own thread, with a command like
./some_test --gtest_filter=Test.Name
Now, some_test also looks for the environment variables GTEST_SHARD_INDEX and GTEST_TOTAL_SHARDS (which is unfortunate, I agree these should be flags, not environment variables).
If GTEST_SHARD_INDEX > 0, when we execute that command, the test thinks that it is part of a swarming environment when only one test is being executed and there is more than one shard, so since it's not the first shard, the test is not executed.

gtest-parallel
ehlemur-zz: Yes, it would, unfortunately.
These environment variables are not set by us, but by the swarming infrastructure team, and I don't think they plan to change it in the near future.
We added this flags as a way to making that more visible.

gtest-parallel
ehlemur-zz: I don't think so.
For example, if options.shard_count = 1 and options.shard_index = -1, then options.shard_count <= options.shard_index (1 <= -1) would be false.

ehlemur-zz:
PTAL

gtest-parallel
kwiberg: OK. I don't particularly like it, but if you must, add code that deletes those two variables from the environment before executing the test binary—subprocess has ways of specifying the environment for the new process.

gtest-parallel
kwiberg: But does gtest-parallel have to look at those variables? Why can't the script simply specify --shard_count=$GTEST_TOTAL_SHARDS etc. when it calls gtest-parallel?

gtest-parallel
kwiberg: Oh, right—it's the test on line 293 that's redundant, but you want a separate test in order to be able to give a good error message. How about this, though? I find this easier to read:

  if not (0 <= options.shard_index < options.shard_count):

gtest-parallel
```diff
@@ -281,6 +290,14 @@ elif options.format == 'filter':
...
+  sys.exit("Invalid shard index: %d. Must be between 0 and %d." %
```
pbos: Can you call parser.error() instead of sys.exit?

gtest-parallel
```diff
@@ -281,6 +290,14 @@ elif options.format == 'filter':
...
+           (options.shard_index, options.shard_count - 1))
+
```
pbos: Remove extra empty line.

gtest-parallel
pbos: Agreed, calling this with environment variables set on Windows/Linux  seems to be not impossible. This is a huge foot-shooter, can you do os.environ.get() in the chromium buildbot script outside that calls the gtest-parallel binary and simply specify those as command-line args?

pbos (changes requested):

gtest-parallel
ehlemur-zz: Yeah. But I can't make those changes. The test is invoked by some swarming logic somewhere else, we just ask swarming to run a certain command in a number of machines, and it prepares the machines and environment variables somehow.
I could open a feature request for the swarming

pbos: Can you have the swarming feature run swarm_gtest_parallel.py instead of gtest-parallel, and have that be a script on top of gtest-parallel that calls it with the swarm flags set from the environment, and just passing forward the other parameters? I'm really reluctant to have the env-dependent defaults, it seems like a bigger foot shooter.

ehlemur-zz:
PTAL

gtest-parallel
```diff
@@ -360,6 +381,7 @@ def run_job((command, job_id, test)):
   with tempfile.NamedTemporaryFile(dir=options.output_dir, delete=False) as log:
     sub = subprocess.Popen(command + ['--gtest_filter=' + test] +
                              ['--gtest_color=' + options.gtest_color],
+                           env=test_env,
```
pbos: This confuses me, why is the test environment modified here? Do the binaries pick up GTEST_SHARD.. otherwise?

gtest-parallel
```diff
@@ -279,7 +288,14 @@ if options.format == 'raw':
...
   logger = FilterFormat()
 else:
-  sys.exit("Unknown output format: " + options.format)
+  parser.error("Unknown output format: " + options.format)
+
+if options.shard_count < 1:
+  parser.error("Invalid number of shards: %d. Must be at least 1." %
+               options.shard_count)
+if not (0 <= options.shard_index < options.shard_count):
+  parser.error("Invalid shard index: %d. Must be between 0 and %d." %
```
pbos: .. and %d (less than the number of shards).

gtest-parallel
ehlemur-zz: Yes. And we don't want that, since that causes the bug I described earlier.

pbos: Can you put a comment above where test_env is created describing this problem?

ehlemur-zz:
I created a feature request [here](https://bugs.chromium.org/p/chromium/issues/detail?id=666335) to use flags instead of env variables.

ehlemur-zz:
PTAL
Got rid of env variables. A buffer script is a good idea.

gtest-parallel
```diff
@@ -266,12 +266,10 @@ parser.add_option('--format', type='string', default='filter',
                   help='output format (raw,filter)')
 parser.add_option('--print_test_times', action='store_true', default=False,
                   help='When done, list the run time of each test')
-parser.add_option('--shard_count', type='int',
-                  default=int(os.environ.get('GTEST_TOTAL_SHARDS', 1)),
+parser.add_option('--shard_count', type='int', default=1,
                   help=('Total number of shards (for sharding test execution '
```
pbos: lowercase help, remove ()s around help and remove default values + "."s

e.g. help='total number of shards (for sharding test execution between multiple machines)'

And

help='zero-indexed number identifying this shard (for sh..'

Can you also change the above help from "When done.." to 'list the run time of each test at the end of execution' ?

ehlemur-zz:
Done :)

gtest-parallel
```diff
@@ -265,14 +265,13 @@ parser.add_option('--gtest_also_run_disabled_tests', action='store_true',
 parser.add_option('--format', type='string', default='filter',
                   help='output format (raw,filter)')
 parser.add_option('--print_test_times', action='store_true', default=False,
-                  help='When done, list the run time of each test')
+                  help='List the run time of each test at the end of execution')
```
pbos: lowercase help here and below

pbos (approved):

pbos:
When this (making help text lowercase) is done, please have @kwiberg finish reviewing and merge it, thanks!

He can also show you how to import to webrtc, specifically updating the README file with the hash. Do this as soon as the buffer script is ready to not break existing runs.

ehlemur-zz:
Done.
So, should I wait for @kwiberg as well?

pbos:
Yes please, since he started commenting.

gtest-parallel
```diff
@@ -279,7 +285,15 @@ if options.format == 'raw':
...
 else:
-  sys.exit("Unknown output format: " + options.format)
+  parser.error("Unknown output format: " + options.format)
+
+if options.shard_count < 1:
+  parser.error("Invalid number of shards: %d. Must be at least 1." %
+               options.shard_count)
+if not (0 <= options.shard_index < options.shard_count):
+  parser.error("Invalid shard index: %d. Must be between 0 and %d "
+               "(less than the number of shards)." %
```
kwiberg, line 295: Maybe drop "less than"? Your call.

kwiberg (approved):
I didn't catch why you removed the environment cleanup again, but as long as it works for you I'm not going to complain. :-)

Peter, when you merge this, the commits will get squashed, right? 7 commits of back-and-forth doesn't make for nice history...

ehlemur-zz:
I think it's fine that way.
The reason for dropping environment variables is that we plan to create a buffer script that converts the environment variables to flags and call gtest-parallel with them.

gtest-parallel
ehlemur-zz: I think it's fine that way

kwiberg:
Yeah, I got that. But I didn't catch why you no longer need to clear GTEST_SHARD_INDEX and GTEST_TOTAL_SHARDS from the environment where you execute the tests. I'm assuming that's also the responsibility of the buffer script, then?

ehlemur-zz:
Yup

ehlemur-zz:
Along with an explanation of why it's needed

PR #13, opened 2016-11-24
add the ability to run test in groups and not as individuals

supershmup:
i had an issue with resources so had to split the tests to groups.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -253,6 +253,8 @@ parser.add_option('-r', '--repeat', type='int', default=1,
                   help='repeat tests')
 parser.add_option('--failed', action='store_true', default=False,
                   help='run only failed and new tests')
+parser.add_option('--group_tests', action='store_true', default=False,
+                  help='run test in groups and not as individuals')
 parser.add_option('-w', '--workers', type='int',
                   default=multiprocessing.cpu_count(),
                   help='number of workers to spawn')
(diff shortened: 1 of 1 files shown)


pbos:
Making an assumption that resources such as servers/files are only shared between test groups seems very specific to your test suite. I'm less inclined to merge it to upstream for that reason since this is more suitable for running self-contained tests. Do feel free to use your patch on top of upstream of course if this assumption holds on your test suite.

WebRTC has modified many of the tests to use tmpfiles or mocks to be parallel-safe and self-contained, and those we couldn't (easily) figure out we run separately as a non-parallel step on our trybots.

(On the off chance that you meant resources such as CPU/memory, then --workers will limit the number of active workers.)

supershmup:
Thanks for the kind reply.
I'll check what is webrtc tomorrow.
The reason i needed to group tests, was that tests used files from sheard folders and under windows i had access issues.

pbos:
WebRTC is why I wrote this script (we needed to run our tests in parallel because some were slow). Can you modify your tests to use temporary files instead of hard coded paths, or open them as read-only? Would that solve it for you?

supershmup:
It seemed eazier to to split the tests than to change my framework, i'll try it today, thank you.

pbos:
If you manage it should parallelize even better, so it'll run faster (unless you're dwarfed by some slow test not in these groups).

PR #14, opened 2016-12-05
Add an option to dump the results of the tests to a JSON machine-readable file.

ehlemur-zz:
Add an option to dump the results of the tests to a JSON machine-readable file.
The format of the file is described in https://www.chromium.org/developers/the-json-test-results-format

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -220,6 +220,8 @@ class CollectTestResults(object):
 class IgnoreTestResults(object):
   def log(self, test, result):
     pass
+  def dump_to_file_and_close(self):
+    pass
 
 # Record of test runtimes. Has built-in locking.
 class TestTimes(object):
@@ -334,9 +336,8 @@ if not (0 <= options.shard_index < options.shard_count):
                "(less than the number of shards)." %
                (options.shard_index, options.shard_count - 1))
 
-test_results = IgnoreTestResults()
-if options.dump_json_test_results:
-  test_results = CollectTestResults(options.dump_json_test_results)
+test_results = (IgnoreTestResults() if options.dump_json_test_results is None
+                else CollectTestResults(options.dump_json_test_results))
 
 # Find tests.
 save_file = os.path.join(os.path.expanduser("~"), ".gtest-parallel-times")
@@ -472,6 +473,5 @@ if options.print_test_times:
               if times.get_test_time(test_binary, test) is not None)
   for (time_ms, test_binary, test) in ts:
     print "%8s %s" % ("%dms" % time_ms, test)
-if options.dump_json_test_results:
-  test_results.dump_to_file_and_close()
+test_results.dump_to_file_and_close()
 sys.exit(-signal.SIGINT if sigint_handler.got_sigint() else exit_code)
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@kwiberg PTAL

ehlemur-zz:
@pbos

gtest-parallel
```diff
@@ -424,4 +470,6 @@ if options.print_test_times:
               if times.get_test_time(test_binary, test) is not None)
   for (time_ms, test_binary, test) in ts:
     print "%8s %s" % ("%dms" % time_ms, test)
+if options.dump_json_test_results:
+  test_results.dump_to_file(options.dump_json_test_results)
```
pbos: Can you hold on to this file handle for the duration of the tests? CollectTestResults() could take dump_json_test_results as a parameter and .dump_to_file could be .write() or something (and the no-op test_results could have an empty .write()).

pbos: Or call it .close() on CollectTestResults, and it writes on close, that makes sense to me. @kwiberg probably has a better method name, but I think CollectTestResults should hold onto the file handle.

ehlemur-zz: What about dump_to_file_and_close()?

ehlemur-zz:
PTAL :)

gtest-parallel
```diff
@@ -424,4 +472,6 @@ if options.print_test_times:
               if times.get_test_time(test_binary, test) is not None)
   for (time_ms, test_binary, test) in ts:
     print "%8s %s" % ("%dms" % time_ms, test)
+if options.dump_json_test_results:
+  test_results.dump_to_file_and_close()
```
pbos: Implement this in the IgnoreTestResults() class too and make it unconditional.

Also add an empty line before this one.

gtest-parallel
```diff
@@ -295,6 +334,10 @@ if not (0 <= options.shard_index < options.shard_count):
                "(less than the number of shards)." %
                (options.shard_index, options.shard_count - 1))
 
+test_results = IgnoreTestResults()
+if options.dump_json_test_results:
+  test_results = CollectTestResults(options.dump_json_test_results)
```
pbos: I think this was better before as CTR(arg) if opt.dump... else ITR().

gtest-parallel
```diff
@@ -186,6 +187,40 @@ class RawFormat:
...
+        results = results.setdefault(name, {})
+      results.update(result)
+
+  def dump_to_file_and_close(self):
+    json.dump(self.test_results, self.json_dump_file)
+    self.json_dump_file.close()
+
+class IgnoreTestResults(object):
+  def log(self, test, result):
+    pass
```
pbos, line 222: Add dump_to_file_and_close(self)

pbos (changes requested):
dump_to_file_and_close sounds great

gtest-parallel
```diff
@@ -334,9 +336,8 @@ if not (0 <= options.shard_index < options.shard_count):
                "(less than the number of shards)." %
                (options.shard_index, options.shard_count - 1))
 
-test_results = IgnoreTestResults()
-if options.dump_json_test_results:
-  test_results = CollectTestResults(options.dump_json_test_results)
+test_results = (IgnoreTestResults() if options.dump_json_test_results is None
+                else CollectTestResults(options.dump_json_test_results))
```
ehlemur-zz, line 340: What about like this, so the line doesn't get too long?

pbos (approved):

pbos:
I want @kwiberg to OK it as well, then we can squash this. :)

Thanks!

kwiberg (approved):
Looks good to me!

PR #16, opened 2017-01-09
Write a minimal JSON dump file at the beginning.

ehlemur-zz:
When the --dump_json_test_results flag is set, write a minimal JSON dump file at the beginning with the 'interrupted' key set to True, so that if the program is interrupted, a valid JSON file is produced.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -190,9 +190,15 @@ class RawFormat:
 class CollectTestResults(object):
   def __init__(self, json_dump_filepath):
     self.test_results_lock = threading.Lock()
-    self.json_dump_file = open(json_dump_filepath, 'w')
+    self.expected_num_tests = -1
+    self.total_num_tests = 0
+    self.json_dump_filepath = json_dump_filepath
+
+    # Write a minimal JSON file with the 'interrupted' key set to true in case
+    # the execution is interrupted.
+    self.json_dump_file = open(self.json_dump_filepath, 'w')
     self.test_results = {
-        "interrupted": False,
+        "interrupted": True,
         "path_delimiter": ".",
         # Third version of the file format. See the link in the flag description
         # for details.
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@kwiberg @pbos

pbos:
I would prefer this in the wrapper code, or for it to be handled in the interrupt handler if partial data + interrupted = True is interesting.

ehlemur-zz:
It can be done in the wrapper code, but I feel that a valid JSON dump file should always be produced when requested.
The problem with doing it in the interrupt handler is, for example, that if I close the terminal window where gtest-parallel is running, then this code doesn't seem to be executed.

pbos:
Is any of the data collected so far useful if the file is interrupted? I don't feel like any program is responsible for outputting anything if you cancel it.

If the data on the runs that have finished are interesting I think that motivates adding the interrupt handler, otherwise I'd prefer it being handled outside. Maybe both if the interrupt handler doesn't reliably kick in, I don't like the idea of creating the bogus file at the top.

@kwiberg can you help look at the interrupt handler and whether or not it actually kicks in when killing the process? Also chime in if my stance on no empty initial file is crazy.

ehlemur-zz:
> I don't feel like any program is responsible for outputting anything if you cancel it.

You might have a point. It's no big deal to do this in the wrapper instead.

> If the data on the runs that have finished are interesting I think that motivates adding the interrupt handler.

I'm not sure if it's used by the dashboard where we report it. It might not be worth logging it.

Finally, I think it's worth looking into when does the interrupt handler doesn't kick in, but this pull request doesn't address this issue.

pbos:
Let's do it that way then. We can reopen this if partial data is useful, I think. @kwiberg please sync with @ehlemur on interrupt-handler issues.

PR #17, opened 2017-01-31
Save test logs to files named after the test name.

ehlemur-zz:
This will make it easier to find and examine the logs of a given test even if it doesn't fail.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -394,7 +394,9 @@ if options.failed:
 tests.sort(reverse=True, key=lambda x: ((1 if x[0] is None else 0), x))
 
 # Repeat tests (-r flag).
+indices = [str(i/len(tests) + 1) for i in range(options.repeat * len(tests))]
 tests *= options.repeat
+tests = zip(tests, indices)
 test_lock = threading.Lock()
 job_id = 0
 logger.log(str(-1) + ': TESTCNT ' + ' ' + str(len(tests)))
@@ -415,16 +417,14 @@ for logfile in os.listdir(options.output_dir):
 # Run the specified job. Return the elapsed time in milliseconds if
 # the job succeeds, or None if the job fails. (This ensures that
 # failing tests will run first the next time.)
-def run_job((command, job_id, test)):
+def run_job((command, job_id, test, test_index)):
   begin = time.time()
 
-  test_name = re.sub(r'\W', '_', test) + '__'
-  with tempfile.NamedTemporaryFile(dir=options.output_dir, delete=False,
-                                   prefix=test_name) as log:
+  test_name = re.sub(r'\W', '_', test) + '-' + test_index + '.log'
+  with open(os.path.join(options.output_dir, test_name), 'w') as log:
     sub = subprocess.Popen(command + ['--gtest_filter=' + test] +
                              ['--gtest_color=' + options.gtest_color],
-                           stdout=log.file,
-                           stderr=log.file)
+                           stdout=log, stderr=log)
     try:
       code = sigint_handler.wait(sub)
     except sigint_handler.ProcessWasInterrupted:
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg

pbos:
How does this behave if two concurrent gtest-parallel instances run with the same output dir?

ehlemur-zz:
Good point.
What about a LOCK file in the output_dir to ensure only one instance is running per output dir?

pbos:
I'm not sure if that's desirable. Also this might conflict with the --repeat flag since this is one log per run? Can you use tempfile.mkstemp with a suffix containing the testname?

ehlemur-zz:
PTAL
I used a prefix instead of a suffix, since I think a prefix is easier to read.
Also, I realized that the second instance of gtest-parallel that runs in the same output dir will erase all existing logs from the first one [(see here)](https://github.com/google/gtest-parallel/blob/master/gtest-parallel#L411)

pbos:
Forgot that we default to a fresh directory for output-dir, I think you can go with your original approach instead then, feel free to ignore the lock-file issue.

ehlemur-zz:
I think having the prefix is a good idea, because of the --repeat flag :)

pbos:
Can it then be testname-1.log? etc?

ehlemur-zz:
Done. Sorry for the 'stuff' commit...

ehlemur-zz:
Friendly ping

gtest-parallel
```diff
@@ -393,7 +395,7 @@ if options.failed:
 tests.sort(reverse=True, key=lambda x: ((1 if x[0] is None else 0), x))
 
 # Repeat tests (-r flag).
-tests *= options.repeat
+tests = list(itertools.product(tests, range(1, options.repeat + 1)))
```
kwiberg: If I'm not mistaken, this is the same as
`  tests = [(t, i + 1) for t in tests for i in range(options.repeat)]
`
Given that itertools.product isn't *that* well-known, it may be better to use the alternative.

gtest-parallel
```diff
@@ -414,14 +416,14 @@ for logfile in os.listdir(options.output_dir):
 # Run the specified job. Return the elapsed time in milliseconds if
 # the job succeeds, or None if the job fails. (This ensures that
 # failing tests will run first the next time.)
-def run_job((command, job_id, test)):
+def run_job((command, job_id, test, test_index)):
   begin = time.time()
 
-  with tempfile.NamedTemporaryFile(dir=options.output_dir, delete=False) as log:
+  test_name = re.sub(r'\W', '_', test) + '-' + str(test_index) + '.log'
```
kwiberg: The meaning of \W is locale-dependent, and quite complex:

> When the LOCALE and UNICODE flags are not specified, matches any non-alphanumeric character; this is equivalent to the set [^a-zA-Z0-9_]. With LOCALE, it will match any character not in the set [0-9_], and not defined as alphanumeric for the current locale. If UNICODE is set, this will match anything other than [0-9_] plus characters classified as not alphanumeric in the Unicode character properties database.

Maybe just say [^a-zA-Z0-9], which is slightly longer but easier to reason about? It would be bad to let stuff like dots and slashes through here...

gtest-parallel
```diff
@@ -447,9 +449,9 @@ def worker():
     job = None
     test_lock.acquire()
     if job_id < len(tests):
-      (_, test_binary, test, command) = tests[job_id]
+      (_, test_binary, test, command), test_index = tests[job_id]
```
kwiberg: Preserve the outer parentheses?
`((_, test_binary, test, command), test_index) = tests[job_id]`

kwiberg (changes requested):
Looks good, but I have a few nits.

Oh, and this should all become a single commit when it lands, I trust?

ehlemur-zz:
PTAL.
I think so, but I'm not sure how to do it.

kwiberg (approved):
Looks good! But you don't need to import itertools anymore.

ehlemur-zz:
Done :)

PR #18, opened 2017-02-06
Add a timeout option to gtest-parallel.

ehlemur-zz:
Stop the tests' execution once the timeout has been exceeded.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -47,7 +47,7 @@ class SigintHandler(object):
     self.__lock = threading.Lock()
     self.__processes = set()
     self.__got_sigint = False
-    signal.signal(signal.SIGINT, self.__sigint_handler)
+    signal.signal(signal.SIGINT, lambda signal_num, frame: self.interrupt())
   def __on_sigint(self):
     self.__got_sigint = True
     while self.__processes:
@@ -55,9 +55,6 @@ class SigintHandler(object):
         self.__processes.pop().terminate()
       except OSError:
         pass
-  def __sigint_handler(self, signal_num, frame):
-    with self.__lock:
-      self.__on_sigint()
   def interrupt(self):
     with self.__lock:
       self.__on_sigint()
@@ -138,6 +135,7 @@ class FilterFormat:
     if test_ids:
       self.out.permanent_line("%s (%s/%s):" %
                               (message, len(test_ids), self.total_tests))
+      test_ids = sorted(test_ids, key=lambda test_id: self.tests[test_id])
       for test_id in test_ids:
         self.out.permanent_line(" %s: %s" % self.tests[test_id])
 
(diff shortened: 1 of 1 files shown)


gtest-parallel
```diff
@@ -58,6 +58,9 @@ class SigintHandler(object):
   def __sigint_handler(self, signal_num, frame):
     with self.__lock:
       self.__on_sigint()
+  def interrupt(self):
```
ehlemur-zz: Maybe we could get rid of __sigint_handler and use this method instead?
Is interrupt a good name?

ehlemur-zz:
@pbos @kwiberg

gtest-parallel
```diff
@@ -465,7 +492,12 @@ def start_daemon(func):
 
 workers = [start_daemon(worker) for i in range(options.workers)]
 
-[t.join() for t in workers]
+try:
+  timeout.start()
```
kwiberg: For paranoia reasons, it's probably better to start the timeout before you start the worker threads. So IOW, maybe move the "workers =" line into the try...except, below this line?

gtest-parallel
```diff
@@ -314,6 +334,9 @@ parser.add_option('--dump_json_test_results', type='string', default=None,
                   help='Saves the results of the tests as a JSON machine-'
                        'readable file. The format of the file is specified at '
                        'https://www.chromium.org/developers/the-json-test-results-format')
+parser.add_option('--timeout', type='int', default=None,
+                  help='Interrupt all remaining processes after a given '
+                       'timeout.')
```
kwiberg: Document what the integer means. Seconds?

gtest-parallel
```diff
@@ -272,6 +291,7 @@ class TestTimes(object):
     except IOError:
       pass  # ignore errors---saving the times isn't that important
 
+
```
kwiberg: Spurious newline?

gtest-parallel
kwiberg: Yes, interrupt is a good name. But __sigint_handler is registered as a sigint handler, so if you eliminated it you'd have to... yes, you could register a lambda as sigint handler instead; the lambda would just ignore the signal_num and frame arguments and call interrupt(). Do that.

gtest-parallel
```diff
@@ -168,11 +183,9 @@ class FilterFormat:
     stdout_lock.release()
 
   def end(self):
-    if self.failures:
-      self.out.permanent_line("FAILED TESTS (%d/%d):"
-                              % (len(self.failures), self.total_tests))
-      for (binary, test) in self.failures:
-        self.out.permanent_line(" " + binary + ": " + test)
+    self.print_tests("FAILED TESTS", self.failures)
+    interruptions = set(self.started).difference(set(self.finished))
```
kwiberg: Can also be written as
`set(self.started) - set(self.finished)`

gtest-parallel
```diff
@@ -168,11 +183,9 @@ class FilterFormat:
...
+    self.print_tests("INTERRUPTED TESTS", interruptions)
```
kwiberg: 1. Don't we already print failed tests?
2. When you print lists of tests, would it make sense to sort them first?

kwiberg (changes requested):

ehlemur-zz:
PTAL :smile:

ehlemur-zz:
I think it's more readable the way it is.

kwiberg (approved):
lgtm

PR #19, opened 2017-03-23
Save logs into passed, failed and interrupted directories.

ehlemur-zz:
This will make it easier to find the failures and crashes, while still saving the passing tests' logs.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -27,8 +27,8 @@ import tempfile
 import thread
 import threading
 import time
-import zlib
 import zipfile
+import zlib
 
 # An object that catches SIGINT sent to the Python process and notices
 # if processes passed to wait() die by SIGINT (we need to look for
@@ -189,6 +189,9 @@ class FilterFormat:
     interruptions = set(self.started) - set(self.finished)
     self.print_tests("INTERRUPTED TESTS", interruptions)
     self.out.flush_transient_output()
+
+    if not self.passed:
+      return
 
     with zipfile.ZipFile(self.zip_file, 'w') as logs_zip:
       for passing in self.passed:
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -28,6 +28,7 @@ import thread
 import threading
 import time
 import zlib
+import zipfile
 
 # An object that catches SIGINT sent to the Python process and notices
 # if processes passed to wait() die by SIGINT (we need to look for
@@ -127,9 +128,10 @@ class FilterFormat:
 
   tests = {}
   outputs = {}
+  passed = []
   started = []
-  finished = []
   failures = []
+  finished = []
 
   def print_tests(self, message, test_ids):
     if test_ids:
@@ -157,7 +159,9 @@ class FilterFormat:
       self.finished.append(job_id)
       (binary, test) = self.tests[job_id]
       self.print_test_status(test, time_ms)
-      if exit_code != 0:
+      if exit_code == 0:
+        self.passed.append(job_id)
+      else:
         self.failures.append(job_id)
         with open(self.outputs[job_id]) as f:
           for line in f.readlines():
@@ -185,6 +189,11 @@ class FilterFormat:
     interruptions = set(self.started) - set(self.finished)
     self.print_tests("INTERRUPTED TESTS", interruptions)
     self.out.flush_transient_output()
+
+    with zipfile.ZipFile(self.zip_file, 'w') as logs_zip:
+      for passing in self.passed:
+        logs_zip.write(self.outputs[passing])
+        os.remove(self.outputs[passing])
 
 class RawFormat:
   def log(self, line):
@@ -346,6 +355,7 @@ if options.format == 'raw':
   pass
 elif options.format == 'filter':
   logger = FilterFormat()
+  logger.zip_file = os.path.join(options.output_dir, 'passing_test_logs.zip')
 else:
   parser.error("Unknown output format: " + options.format)
 
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -128,6 +128,7 @@ class FilterFormat:
 
   tests = {}
   outputs = {}
+  passed = False
   failed = []
   started = set()
 
@@ -161,6 +162,7 @@ class FilterFormat:
       (binary, test) = self.tests[job_id]
       self.print_test_status(test, time_ms)
       if exit_code == 0:
+        self.passed = True
         self.move_to('passed', file_name)
       else:
         with open(file_name) as f:
@@ -179,8 +181,14 @@ class FilterFormat:
   def end(self):
     if self.failed:
       self.print_tests("FAILED TESTS", self.failed)
+    else:
+      os.rmdir(os.path.join(self.output_dir, 'failed'))
     if self.started:
       self.print_tests("INTERRUPTED TESTS", self.started)
+    else:
+      os.rmdir(os.path.join(self.output_dir, 'interrupted'))
+    if not self.passed:
+      os.rmdir(os.path.join(self.output_dir, 'passed'))
     self.out.flush_transient_output()
 
 class RawFormat:
@@ -453,7 +461,6 @@ def run_job((command, job_id, test, test_index)):
     except sigint_handler.ProcessWasInterrupted:
       thread.exit()
     runtime_ms = int(1000 * (time.time() - begin))
-    logger.logfile(job_id, log.name)
 
   test_results.log(test, {
       "expected": "PASS",
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -20,6 +20,7 @@ import multiprocessing
 import optparse
 import os
 import re
+import shutil
 import signal
 import subprocess
 import sys
@@ -27,7 +28,6 @@ import tempfile
 import thread
 import threading
 import time
-import zipfile
 import zlib
 
 # An object that catches SIGINT sent to the Python process and notices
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@kwiberg @pbos 
Please take a look :smile:

gtest-parallel
```diff
@@ -28,6 +28,7 @@ import thread
 import threading
 import time
 import zlib
+import zipfile
```
kjellander-unity: sort alphabetically.

gtest-parallel
```diff
@@ -186,6 +190,11 @@ class FilterFormat:
     self.print_tests("INTERRUPTED TESTS", interruptions)
     self.out.flush_transient_output()
 
+    with zipfile.ZipFile(self.zip_file, 'w') as logs_zip:
```
kjellander-unity: Please handle the case where self.passed = []
No need to write an empty zip file then.

kjellander-unity (commented):
I'd like to know some kind of ballpark figure on how long this takes on, let's say 3000 small text files.

ehlemur-zz:
I ran modules_unittests, archiving 2746 test logs took about ~1s extra.

kjellander-unity (approved):
I'd prefer to not use return, but it's not the end of the world.

gtest-parallel
```diff
@@ -346,6 +358,7 @@ if options.format == 'raw':
   pass
 elif options.format == 'filter':
   logger = FilterFormat()
+  logger.zip_file = os.path.join(options.output_dir, 'passing_test_logs.zip')
```
kwiberg: This zip file will be created in interactive runs too—maybe not what we want. How about adding an option, e.g. --save_passing_test_logs=foo.zip, that sets the name of the zip file, and causes no zip file to be created if it isn't set?

gtest-parallel
```diff
@@ -186,6 +190,14 @@ class FilterFormat:
     self.print_tests("INTERRUPTED TESTS", interruptions)
     self.out.flush_transient_output()
 
+    if not self.passed:
+      return
+
```
kwiberg: 1. You can save a bunch of lines if you invert this to "if self.passed:".
2. Why don't you want an empty zip file? Won't that just require more logic in whatever script is calling us?

kwiberg (dismissed):

gtest-parallel
kjellander-unity: How about changing this to create one single log file instead of the zipping, from all job IDs (passing and failing ones). Then maybe also emit separate log files for the interrupted/failed tests.
That would fit the interactive use case as well (assuming people prefer one log file rather than thousands of files). It will also solve the problem on the bot (that we archive thousands of files per test run).

ehlemur-zz: One log for both interrupted and failed, or two separate logs?

kjellander-unity: Let's have one for everything, then separate logs for failed and interrupted. How about this style:
modules_unittests.log (contains all logs)
modules_unittests_FAILED.log (only failed)
modules_unittests_INTERRUPTED.log (only interrupted).

ehlemur-zz:
PTAL

pbos:
This issue title confuses me now. Suggesting something like Add an option for outputting zipped logs.

Also, can we consider storing the logs as individual files (passed/foobinary.footest.run-1.txt) in passed, failed, interrupted directories inside the zip? The bot can concatenate them fine, but splitting them up would be hard.

kjellander-unity:
We cancelled the idea of zipped logs. I think a full log but one file instead of thousands is useful. The amount of space that text file occupies is not a problem. Zipping just adds more complexity and is more annoying to browse.

gtest-parallel
```diff
@@ -181,9 +188,16 @@ class FilterFormat:
     stdout_lock.release()
 
   def end(self):
-    self.print_tests("FAILED TESTS", self.failures)
-    interruptions = set(self.started) - set(self.finished)
-    self.print_tests("INTERRUPTED TESTS", interruptions)
+    if self.failures:
+      self.print_tests("FAILED TESTS", self.failures)
+      self.write_logs_to_file('Failed tests.log', self.failures)
```
kjellander-unity: Let's not use spaces in the filename, that's never a good idea. I'd also prefer all lowercase.

gtest-parallel
```diff
@@ -27,6 +27,7 @@ import tempfile
 import thread
 import threading
 import time
+import zipfile
```
kjellander-unity: No longer used

pbos:
Is it feasible to do this concatenation outside the script or do you need this in here? It seems very specific to WebRTC.

gtest-parallel
```diff
@@ -127,17 +128,26 @@ class FilterFormat:
...
   failures = []
+  finished = []
+
+
+  def write_logs_to_file(self, bundle_name, tests_ids):
+    with open(os.path.join(self.logs_dir, bundle_name), 'w') as logs_bundle:
+      for test_id in tests_ids:
+        log_file = self.outputs[test_id]
+        test_name = os.path.basename(log_file)
+        logs_bundle.write('TEST: ' + test_name + '\n\n')
```
kjellander-unity: Do we need this extra information? I'd prefer if the log looks exactly like a regular gtest output, like the test would have been executed without gtest-parallel.

gtest-parallel
```diff
@@ -127,17 +128,26 @@ class FilterFormat:
...
+        with open(self.outputs[test_id]) as test_log:
+          logs_bundle.write(test_log.read())
+        logs_bundle.write('\n\n')
 
   def print_tests(self, message, test_ids):
```
kjellander-unity: I don't think we need the message. Can't we just make it simpler and just output the test output? The filename showcases what the content represents (and since the tests will only show RUN and not OK for interrupted tests, I think everyone will understand).

kjellander-unity:
I don't think anyone is interested in having thousands of log files created when running a large test. How can that be specific to WebRTC? Since this script wraps each test execution we'd have to add another layer of wrapper script to do the concatenation. We can't do it after the run, as we specify the output folder as the swarming output, which is automatically archived as soon the swarming script finishes.

pbos:
Do you actually want these as separate logfile outputs or would you like the test tool to optionally also dump stdio of passing tests to stdout?

ehlemur-zz:
We're interesting in dumping them to a file, not to stdout.

pbos:
Would it be significantly different from capturing stdout?

ehlemur-zz:
You mean like dumping it to stdout and then capturing stdout to a file in a wrapper script?
It'd be more difficult, since we'd also capture the messages that gtest-parallel normally prints and that we want to show to the user.

And just dumping it to stdout, without capturing it later to a file, will make it very hard to see the failures and other messages that gtest-parallel might print.

We could make this an optional flag, which if set, would classify and concatenate the logs into the three files.

pbos:
Yeah, that's fine with me, but I think the feature overlaps with `output_dir` and the concatenation can be done outside like this:

Can we instead update the `--output_dir` option to output logs under `passed/`, `failed/` and `interrupted/` subdirectories (move them once the test is complete) instead of adding a new option? Your calling script can `cat output_dir/passed/* > my_passed_tests.log` to get them into one file.

ehlemur-zz:
Ok, SGTM

gtest-parallel
```diff
@@ -169,8 +181,8 @@ class FilterFormat:
       self.total_tests = int(arg.split(' ', 1)[1])
       self.out.transient_line("[0/%d] Running tests..." % self.total_tests)
 
-  def logfile(self, job_id, name):
-    self.outputs[job_id] = name
+  def logfile(self, job_id, file_name):
```
pbos: If this one is not needed here, can you work it away for the RawFormat class too and get rid of the call here?

ehlemur-zz:
What about something like this?

gtest-parallel
```diff
@@ -420,30 +419,32 @@ tests.sort(reverse=True, key=lambda x: ((1 if x[0] is None else 0), x))
...
 
 exit_code = 0
 
+# Remove files from old test runs.
+if os.path.isdir(options.output_dir):
+  shutil.rmtree(options.output_dir)
 # Create directory for test log output.
 try:
-  os.makedirs(options.output_dir)
+  for result in 'passed', 'failed', 'interrupted':
```
pbos, line 438: I think these should only be made if there are any failed/interrupted tests, so you can do this in the code that moves the logs.

gtest-parallel
```diff
@@ -127,75 +128,72 @@ class FilterFormat:
...
 class RawFormat:
-  def log(self, line):
+  def log(self, job_id, command, args=tuple(), file_name=None):
     stdout_lock.acquire()
+    line = "%d: %s %s" % (job_id, command, ' '.join(str(arg) for arg in args))
     sys.stdout.write(line + "\n")
+    if command == "EXIT":
+      with open(file_name) as f:
+        for line in f:
+          sys.stdout.write("%d > %s" % (job_id, line))
```
pbos, line 201: Does this print the last \n? I'm not sure if the last line contains a \n?

gtest-parallel
```diff
@@ -127,75 +128,72 @@ class FilterFormat:
...
     stdout_lock.release()
 
   def end(self):
-    self.print_tests("FAILED TESTS", self.failures)
-    interruptions = set(self.started) - set(self.finished)
-    self.print_tests("INTERRUPTED TESTS", interruptions)
+    if self.failed:
+      self.print_tests("FAILED TESTS", self.failed)
+    if self.started:
+      self.print_tests("INTERRUPTED TESTS", self.started)
```
pbos: Can we remove the interrupted dir if it's empty here?

gtest-parallel
pbos: Never mind, see other comment about removing empty dirs at the end of run.

gtest-parallel
pbos: Actually, removing all empty dirs here would be a good way to do it instead.

pbos:
I think this looks good overall. Thanks for making these changes!

gtest-parallel
ehlemur-zz: It does. It displays the output correctly.

gtest-parallel
ehlemur-zz: Done :smile:

ehlemur-zz:
PTAL :smile:

gtest-parallel
```diff
@@ -127,75 +128,80 @@ class FilterFormat:
...
 
   def print_test_status(self, last_finished_test, time_ms):
     self.out.transient_line("[%d/%d] %s (%d ms)"
                             % (self.finished_tests, self.total_tests,
                                last_finished_test, time_ms))
 
-  def handle_meta(self, job_id, args):
-    (command, arg) = args.split(' ', 1)
+  def log(self, job_id, command, args=tuple(), file_name=None):
+    stdout_lock.acquire()
```
pbos: Should this whole thing be `with stdout_lock:` instead of manual acquire/release?

pbos:
Removing @kwiberg's requested changes since they only applied to the .zip option.

gtest-parallel
ehlemur-zz: Yeah, that's better.

ehlemur-zz: Done.

pbos:
LGTM, thanks.

PR #22, opened 2017-04-20
Refactor gtest-parallel.

ehlemur-zz:
What I tried to do here is to create some classes to store the information about the tests and test executions so they're not saved in multiple times, and move things from the global scope to functions to improve readability.
I also introduced the capability to retry failed tests.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -30,6 +30,10 @@ import threading
 import time
 import zlib
 
+
+INF = float("inf")
+
+
 # An object that catches SIGINT sent to the Python process and notices
 # if processes passed to wait() die by SIGINT (we need to look for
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg 
Please take a look.

gtest-parallel
```diff
@@ -30,6 +30,10 @@ import threading
 import time
 import zlib
 
+
+INF = float("inf")
```
pbos, line 34: Can you use None instead?

pbos:
Is it possible for you to break this up into smaller changes? The linediff is huge, and you're making the file about 20% larger.

ehlemur-zz:
I'll try.
If it's not too complex, can you take an overall view of the changes and tell me what you think?

PR #24, opened 2017-04-27
Fix errors introduced in the last pull request.

ehlemur-zz:
We were using exit_code instead of task.exit code, and broke the print_test_times option.
Also, this is might be related to [this issue](https://bugs.chromium.org/p/webrtc/issues/detail?id=7524).


ehlemur-zz:
@pbos @kwiberg

PR #26, opened 2017-05-02
Add TaskManager class and support to repeat failed tests.

ehlemur-zz:
TaskManager is responsible for dealing with times, logger and test_results, keeping track of what tasks have been executed and which tasks should be repeated.
The only semantic change should be that now, when we repeat the tasks with the --repeat flag, we wait until we finish executing the test to schedule the next execution. This is because running several instances of a test at the same time might cause problems.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import cPickle
+import collections
 import errno
 import gzip
 import json
@@ -117,10 +118,10 @@ class Outputter(object):
 
 
 class Task(object):
-  def __init__(self, task_id, test_binary, test_name, test_command,
-               execution_number, last_execution_time, output_dir):
-    self.task_id = task_id
+  def __init__(self, test_binary, test_name, test_command, execution_number,
+               last_execution_time, output_dir):
     self.test_name = test_name
+    self.output_dir = output_dir
     self.test_binary = test_binary
     self.test_command = test_command
     self.execution_number = execution_number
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg 
Please take a look :)

pbos:
--repeat needs to support running the same test in parallel, --gtest_filter=OneTest.OneTest -r2000 should run that same test in parallel.

ehlemur-zz:
Ok, done.
Retries are done immediately, in the same thread.

gtest-parallel
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import cPickle
+import collections
```
kwiberg: This appears to be unused.

gtest-parallel
```diff
@@ -117,19 +118,23 @@ class Outputter(object):
 
 
 class Task(object):
```
kwiberg, line 119: Could you add a comment describing what this class does? (It should already have been there. This patch just makes the problem a bit worse.)

gtest-parallel
```diff
@@ -155,70 +159,102 @@ class Task(object):
...
-    test_results.log(self.test_name, {
-        "expected": "PASS",
-        "actual": "PASS" if self.exit_code == 0 else "FAIL",
-        "time": self.runtime_ms,
-    })
-    logger.log_exit(self.task_id)
 
 
-stdout_lock = threading.Lock()
+class TaskManager(object):
```
kwiberg, line 172: And a comment for this class too?

kwiberg (changes requested):
Looks good.

ehlemur-zz:
Please take a look.

kwiberg (approved):

PR #27, opened 2017-05-03
Test that test_results is not nome before calling dump_to_file_and_close

ehlemur-zz:
I need to be more careful 😢 
And we need some tests.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -575,7 +575,8 @@ if task_manager.started:
 
 logger.flush()
 times.write_to_file(save_file)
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg 
Please take a look.

kwiberg (approved):

kwiberg:
Yup, some tests would definitely be good. Shouldn't even be that hard to write—construct a fake test binary (in Python!) that fails and succeeds as necessary, and you're golden.

PR #28, opened 2017-05-04
Fix global_exit_code.

ehlemur-zz:
There were two bugs in the code:
- global_exit_code was always set to task.exit_code, so that if a passing task was executed after a failing task, global_exit_code was reset to 0.
- If we are in the last iteration on the loop, and the task didn't succeed, we still create a new task. Then, we set global_exit_code to task.exit_code, which is None.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -217,7 +217,7 @@ class TaskManager(object):
         self.failed.append(task)
 
   def run_task(self, task):
-    for _ in range(options.retry_failed + 1):
+    for try_number in range(options.retry_failed + 1):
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg 
Yet another bug.
Please take a look.

kwiberg (approved):
lgtm. And as you said before, the lack of tests is getting painful. If you're going to keep making changes, consider building a tiny test suite that verifies that a few common cases work.

ehlemur-zz:
Yes, I'll start working on adding tests.

PR #29, opened 2017-05-04
Move gtest-parallel to gtest_parallel.py and make gtest-parallel a thin wrapper.

ehlemur-zz:
This will allow us to import it as a library and add tests.


ehlemur-zz:
@pbos @kwiberg 
Please take a look.
Don't know if this is the best way to introduce tests. I'd like to have unit tests, not *only* testing the whole script.

pbos:
This diff is nasty looking, is there any way for it to detect renames during uploading or similar so that it can easily be reviewed?

ehlemur-zz:
Maybe I can temporarily copy gtest_parallel.py into gtest-parallel?
I don't know how to tell git that we renamed the file, since the original one still exists.

ehlemur-zz:
Or maybe rename gtest-parallel to something else in the meantime?

pbos:
you can do everything except the rename during review, just call gtest-parallel gtest-herkaderk meanwhile?

ehlemur-zz:
Ok, please take a look now :)

gtest_parallel.py
```diff
@@ -382,25 +385,26 @@ def write_to_file(self, save_file):
       pass  # ignore errors---saving the times isn't that important
 
 
-def find_tests(binaries, additional_args):
-  test_count = 0
+def find_tests(binaries, run_disabled_tests, only_failed_and_new, gtest_filter,
+               gtest_color, shard_index, shard_count, repeat, output_dir,
+               additional_args, times, run_command_fn):
```
kwiberg: This argument list is... impressive. Maybe consider passing an options object instead? (Because of the duck typing, callers can easily make up a suitable object that has nothing to do with argparse.)

As an added bonus, this would make the diff smaller.

kwiberg (changes requested):
Looks good, but may I suggest splitting it up into two separate pull requests: one that moves all the code to a .py file (including moving all the top-level stuff into functions) and makes gtest-parallel a thin wrapper, and one that adds the unit test file and makes the changes required to make that work.

If you feel that this is a bad idea, I won't insist.

pbos:
I'd agree with that split, it shouldn't be too tricky and makes sense.

ehlemur-zz:
I did that, but for some reason I can't get it to work.
It gets stuck at the `task = subprocess.Popen(self.test_command, stdout=log, stderr=log)` line.

ehlemur-zz:
I had to add a `main()` function, for some reason.
Other than moving all the top level code into main, and declaring options, times, logger, task_manager and timeout as global, there shouldn't be more changes.
On the other hand, how important is to keep gtest-parallel as opposed to renaming it to gtest_parallel.py?

kwiberg (changes requested):
I *think* this is good, but like before the diff is impossible to read. Would you please use the same trick as before? (Name the wrapper gtest-parallel2 so that git realizes that you've renamed gtest-parallel to gtest_parallel.py. And remove the trick in a separate commit at the end, which we can choose to not see while reviewing.)

As for your question: No, I don't think we need gtest-parallel; just gtest_parallel.py should be enough. But you probably need to make that change in a separate pull request later, after our users (just WebRTC, I guess?) have had time to switch.

pbos:
As one of our users, I'd like us to keep the source-code-language agnostic gtest-parallel alias. :)

kwiberg:
Well, it's very cheap to keep the executable "gtest-parallel" around, so fine by me.

ehlemur-zz:
Ok :)
PTAL

pbos:
GitHub is terrible at these diffs, use this one:

https://pastebin.mozilla.org/9021221

gtest_parallel.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python2
-# Copyright 2013 Google Inc. All rights reserved.
+# Copyright 2017 Google Inc. All rights reserved.
```
pbos: I don't think this should be changed. Semantically this is the same file, so do 2013.

gtest-parallel2
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python2
+# Copyright 2013 Google Inc. All rights reserved.
```
pbos: This file is new but with the old name, do 2017 here.

gtest_parallel.py
```diff
@@ -467,119 +467,130 @@ def start_daemon(func):
...
-
-logger.flush()
-times.write_to_file(save_file)
-if test_results:
-  test_results.dump_to_file_and_close()
-
-if sigint_handler.got_sigint():
-  task_manager.global_exit_code = -signal.SIGINT
-sys.exit(task_manager.global_exit_code)
+options = None
```
pbos: Just double checking, are these polluting the global namespace of anything importing this file?

pbos (changes requested):

gtest_parallel.py
```diff
@@ -467,119 +467,130 @@ def start_daemon(func):
...
+  logger.flush()
+  times.write_to_file(save_file)
+  if test_results:
+    test_results.dump_to_file_and_close()
+
+  if sigint_handler.got_sigint():
+    task_manager.global_exit_code = -signal.SIGINT
+  sys.exit(task_manager.global_exit_code)
+
+if __name__ == '__main__':
```
pbos: Suggesting that you remove this __main__ check in here and only use gtest-parallel as the launcher. Up to you though.

gtest_parallel.py
pbos: Also, how about setting these in main(), so if you run main() twice that still works?

ehlemur-zz: They need to be global, they're used as global variables in other places.
I wanted to make this CL just about moving gtest-parallel to gtest_parallel.py and leaving gtest-parallel as a thin wrapper. I had to introduce a main function for some reason.
So, should I also fix these global invocations?

ehlemur-zz:
PTAL :)

pbos:
lgtm, @kwiberg can you take a look as well? Do you think the change getting rid of global variables was worth it? I mostly thought that they'd be set inside main() and not once when loading the file.

pbos:
Please update the pull request name/title as well. :)

gtest_parallel.py
kwiberg: They're not polluting the namespace of anything that imports this file, because files in Python are modules, and "global" variables are scoped to the module they live in.

That said, I still agree that these ought to not be module global, but scoped to main(). The way you fixed it in the follow-up commit is good.

gtest_parallel.py
kwiberg: I agree. We don't need two ways to start the program.

kwiberg (approved):

kwiberg:
As far as I'm concerned, this can be merged as soon as gtest-parallel2 is re-renamed.

ehlemur-zz:
Re-renamed :)

kwiberg (approved):
lgtm!

PR #31, opened 2017-05-11
Add tests to gtest-parallel.

ehlemur-zz:


ehlemur-zz:
@pbos @kwiberg 
PTAL.

pbos:
Can you add these tests one at a time so they're easier to review? :)

ehlemur-zz:
Ok :)
PTAL

gtest_parallel.py
```diff
@@ -177,11 +177,12 @@ class TaskManager(object):
   Logger, TestResults and TestTimes classes, and in case of failure, retries the
   test as specified by the --retry_failed flag.
   """
-  def __init__(self, times, logger, test_results, times_to_retry,
+  def __init__(self, times, logger, test_results, task_factory, times_to_retry,
```
pbos, line 180: Is this a factory, or should this be called task_class? Python kinda makes this the same thing. WDYT?

gtest_parallel_tests.py
```diff
@@ -0,0 +1,134 @@
+import collections
```
pbos, line 15: Add license header w/ 2017.

gtest_parallel.py
pbos: Ignore this, I think it's fine. :)

kwiberg (changes requested):
Looks good, except the copyright year as pbos@ said.

ehlemur-zz:
PTAL :)

PR #32, opened 2017-05-15
Add --tests_times_output_dir option to override default location for …

volo-zyko:
….gtest-parallel-times.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -567,7 +567,10 @@ def main():
     default_cache_path = os.path.join(os.path.expanduser('~'), '.cache')
     cache_path = os.environ.get('XDG_CACHE_HOME', default_cache_path)
 
-  save_file = os.path.join(cache_path, 'gtest-parallel')
+  if os.path.isdir(cache_path):
+    save_file = os.path.join(cache_path, 'gtest-parallel')
+  else:
+    save_file = os.path.join(os.path.expanduser('~'), '.gtest-parallel-times')
 
   times = TestTimes(save_file)
   logger = FilterFormat(options.output_dir)
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -518,10 +518,6 @@ def main():
   parser.add_option('--timeout', type='int', default=None,
                     help='Interrupt all remaining processes after the given '
                          'time (in seconds).')
-  parser.add_option('--tests_times_output_dir', type='string',
-                    default=os.path.expanduser('~'),
-                    help='output directory for saving results of previous '
-                         'tests\' runs')
 
   (options, binaries) = parser.parse_args()
 
@@ -563,8 +559,16 @@ def main():
   if options.dump_json_test_results is not None:
     test_results = CollectTestResults(options.dump_json_test_results)
 
-  save_file = os.path.join(options.tests_times_output_dir,
-                           '.gtest-parallel-times')
+  if sys.platform == 'win32':
+    default_cache_path = os.path.join(os.path.expanduser('~'), 'AppData', 'Local')
+    cache_path = os.environ.get('LOCALAPPDATA', default_cache_path)
+  else:
+    # We don't use xdg module since it's not a standard.
+    default_cache_path = os.path.join(os.path.expanduser('~'), '.cache')
+    cache_path = os.environ.get('XDG_CACHE_HOME', default_cache_path)
+
+  save_file = os.path.join(cache_path, 'gtest-parallel')
+
   times = TestTimes(save_file)
   logger = FilterFormat(options.output_dir)
 
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -77,6 +77,7 @@ class SigintHandler(object):
     return code
 sigint_handler = SigintHandler()
 
+
 # Return the width of the terminal, or None if it couldn't be
 # determined (e.g. because we're not being run interactively).
 def term_width(out):
@@ -91,6 +92,7 @@ def term_width(out):
     return int(out.split()[1])
   except (IndexError, OSError, ValueError):
     return None
+
 
 # Output transient and permanent lines of text. If several transient
 # lines are written in sequence, the new will overwrite the old. We
@@ -114,6 +116,24 @@ class Outputter(object):
   def permanent_line(self, msg):
     self.flush_transient_output()
     self.__out_file.write(msg + "\n")
+
+
+def get_save_file_path():
+  """Return path to file for saving transient data."""
+  if sys.platform == 'win32':
+    default_cache_path = os.path.join(os.path.expanduser('~'),
+                                      'AppData', 'Local')
+    cache_path = os.environ.get('LOCALAPPDATA', default_cache_path)
+  else:
+    # We don't use xdg module since it's not a standard.
+    default_cache_path = os.path.join(os.path.expanduser('~'), '.cache')
+    cache_path = os.environ.get('XDG_CACHE_HOME', default_cache_path)
+
+  if os.path.isdir(cache_path):
+    return os.path.join(cache_path, 'gtest-parallel')
+  else:
+    sys.stderr.write('Directory {} does not exist'.format(cache_path))
+    return os.path.join(os.path.expanduser('~'), '.gtest-parallel-times')
 
 
 class Task(object):
@@ -559,18 +579,7 @@ def main():
   if options.dump_json_test_results is not None:
     test_results = CollectTestResults(options.dump_json_test_results)
 
-  if sys.platform == 'win32':
-    default_cache_path = os.path.join(os.path.expanduser('~'), 'AppData', 'Local')
-    cache_path = os.environ.get('LOCALAPPDATA', default_cache_path)
-  else:
-    # We don't use xdg module since it's not a standard.
-    default_cache_path = os.path.join(os.path.expanduser('~'), '.cache')
-    cache_path = os.environ.get('XDG_CACHE_HOME', default_cache_path)
-
-  if os.path.isdir(cache_path):
-    save_file = os.path.join(cache_path, 'gtest-parallel')
-  else:
-    save_file = os.path.join(os.path.expanduser('~'), '.gtest-parallel-times')
+  save_file = get_save_file_path()
 
   times = TestTimes(save_file)
   logger = FilterFormat(options.output_dir)
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -518,6 +518,10 @@ def main():
   parser.add_option('--timeout', type='int', default=None,
                     help='Interrupt all remaining processes after the given '
                          'time (in seconds).')
+  parser.add_option('--tests_times_output_dir', type='string',
+                    default=os.path.expanduser('~'),
+                    help='output directory for saving results of previous '
+                         'tests\' runs')
 
   (options, binaries) = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


volo-zyko:
I signed it!

volo-zyko:
BTW, I don't mind renaming the command line option to whatever you think appropriate. I just want some way to override location for .gtest-parallel-times file.

pbos:
This sounds like the kind of thing I'd like to not expose an option for.

@kwiberg Is there another appropriate directory that we can use for gtest-parallel data if this is polluting user directories? I'm thinking something like ~/.config/gtest-parallel, if ~/.config/ exists.

@volo-zyko Would placing these runtimes in .config/gtest-parallel work for you?

volo-zyko:
@pbos What if I have a number of independent builds for the same project (say, on different branches) and I don't want to intermingle results for them?

~/.config/gtest-parallel is definitely better than ~/.gtest-parallel-times but still why not allow user to change this default behavior...

kwiberg (changes requested):
I'm fine with this sort of option in general (though I won't approve it if pbos@ still objects), but I'd rather have it specify the full path including file name, rather than just the directory, since that's more flexible

kwiberg:
Yes, ~/.config/gtest-parallel might be better than the current default... hmm... except this isn't a config file, but rather a cache. So maybe not.

In any case, adding a flag like this and changing the default are two separate things, so I don't think this pull request should attempt to do both.

pbos:
Maybe ~/.cache/gtest-parallel is better then (didn't think of/know of that directory).

I think we should use and respect $XDG_CACHE_HOME (use ${XDG_CACHE_HOME}/gtest-parallel) by default, and default to $HOME/.cache/gtest-parallel when it's not available per the XDG spec, then the user can override the cache directory through XDG_CACHE_HOME=foo gtest-parallel rather than rolling our own flags for this behavior.

pbos:
See: https://wiki.archlinux.org/index.php/XDG_Base_Directory_support#User_directories

kwiberg:
Sounds good to me. (But personally I'm still OK with adding a flag. I'll leave it to the two of you to hash out if we should have a flag, $XDG_CACHE_HOME, or both.)

pbos:
I think respecting XDG_CACHE_HOME provides value in itself, and with that specified I don't think the flag is necessary (the bike shed should definitely be red, and have no white frills).

volo-zyko:
Well, I'm fine with using XDG_CACHE_HOME. Anyway allowing both command line option and environment variable can be confusing. I'll try to send another pull request.

volo-zyko:
But after some thinking... Employing $XDG_CACHE_HOME can probably create problems if some test (run by gtest-parallel) changes its behavior depending on this variable. What do you think?

pbos:
.. cross that bridge when we get there, or advocate that the test run by gtest-parallel should also respect XDG_CACHE_HOME.

volo-zyko:
Also, shall we break with this change previous behavior? I.e. if XDG_CACHE_HOME is not set should we use the default $HOME/.cache and thus abandon ~/.gtest-parallel-times at all?

volo-zyko:
> .. cross that bridge when we get there, or advocate that the test run by gtest-parallel should also respect XDG_CACHE_HOME.

You know, we can speculate even further. Say, there are 2 scenarios for running test: 1) manually (in command line or debugger); 2) using build system (something like "make run_tests"). In the first scenario user most like will forget/not know about XDG_CACHE_HOME. In the second case we definitely want to run tests faster and we use gtest-parallel in build scripts and since we don't want to pollute home directory for arbitrary user we run it with XDG_CACHE_HOME. And now if test uses anything from XDG_CACHE_HOME then we can obtain different behaviors depending on how user runs test. So, in some rare cases this can confuse people...

There is no such problem with command line argument. It just cannot change behavior of sub-processes spawned by gtest-parallel. And unless user is interested in tests execution times from previous runs he/she will not see any difference whatsoever.

pbos:
I don't think an ordinary user should have to know that this cache file exists at all, so I think 1) is mostly a non-concern for me. If a normal user needs to care we have bigger problems. I think 2) should work for most systems by just not overriding XDG_CACHE_HOME at all. This should really only affect you if the debug version of a test runs slower than release (in a way that makes release/debug orderings of test times different). I'd argue that this is a corner case, and the cost of getting this wrong is not that high.

I see two original issues, let me know if that's wrong:

1) We pollute the user directory and don't respect XDG_CACHE_HOME.
2) Results are intermingled between different runtimes of invocations of tests with the same name.

I buy 1), and would definitely accept patches for it. We should be able to pollute .cache/gtest-parallel (that's what .cache is for).

I really don't care much for fixing 2) here, unless you're really hitting it *and* it's really hurting you badly. I don't like adding config options that are generally not required or otherwise not well motivated. The tests should still still run fine just not with optimal total runtime (in corner cases) since the estimated total runtimes will be off (also happens for any invocations where you actually update the tests). gtest-parallel still provides a significant speedup vs. non-gtest-parallel, and your times between runs are probably still an improvement under just random ordering.

Is there a third where running parallel invocations of gtest-parallel fail because of contention on the cache file? I'm running two parallel invocations of gtest-parallel right now, and it seems to work fine.

volo-zyko:
Basically, the main thing that I want to fix is the first issue. I.e. avoid polluting user's home directory. Anything else was just an attempt to foresee problems that might create the discussed solution.

pbos:
Thanks, this looks good to me. @ehlemur can you suggest how to best add tests for this?

gtest_parallel.py
```diff
@@ -559,7 +559,16 @@ def main():
...
-  save_file = os.path.join(os.path.expanduser("~"), ".gtest-parallel-times")
+  if sys.platform == 'win32':
+    default_cache_path = os.path.join(os.path.expanduser('~'), 'AppData', 'Local')
+    cache_path = os.environ.get('LOCALAPPDATA', default_cache_path)
+  else:
+    # We don't use xdg module since it's not a standard.
+    default_cache_path = os.path.join(os.path.expanduser('~'), '.cache')
+    cache_path = os.environ.get('XDG_CACHE_HOME', default_cache_path)
+
+  save_file = os.path.join(cache_path, 'gtest-parallel')
```
pbos: Does this create directories here if they are not available (say, you remove .cache or provide a new XDG_CACHE_HOME?)

volo-zyko: Initially I added this code but it looks like this directory has to be created with special permissions. On my machine it's 0700 but I couldn't find exact value in the XDG specification (what about sticky bit?). Also, there is other code which assumes that user's home directory always exists. I understand that there is a huge difference between home and cache but I thought that it's safer to assume it exists. :)

So, I missed that now we can set XDG_CACHE_HOME to an arbitrary value and thus we have to ensure it exists.

volo-zyko: After some more thinking... Since you don't want to exhibit this cache to end user at all then XDG_CACHE_HOME should not be used for managing "gtest-parallel" cache. If XDG_CACHE_HOME is changed for any reason we obey it but not more than that. We should not assume that it's changed for gtest_parallel.py exclusively and create this directory if it doesn't exist. Another reason is that changing of LOCALAPPDATA on Windows is not that easy and normally applications don't change it. So, probably we should not create LOCALAPPDATA if it doesn't exist too.

pbos: I think my main concern was if .cache/ is not available (I don't think this is as standard as having a ~/ directory). Could it fall back to ~/.gtest-parallel-times as a last resort, or ignore the file and spit out some warning?

pbos: I'm fine with a warning here, saying cache directory <fullpath> doesn't exist would be very helpful to print as an error message, just make sure the script doesn't crash because of it.

ehlemur-zz:
This would be much easier to test if we put this in a function. Then maybe we can mock `os` to test it.

pbos:
Looks good to me implementation wise. I agree with stubbing it out to a function (get_save_file_path() maybe) since that's easier to test. Can you add test coverage in gtest_parallel_tests.py?

I'd like for this test 5 paths: On win32, the default path is the ~/AppData/Local/ (1), but LOCALAPPDATA is respected (2).
On non-win32 the default path is  .cache/ (3), but XDG_.. is respected (4).
When this path doesn't exist, ~/.gtest-parallel-times is used (5).

gtest_parallel_tests.py
```diff
@@ -143,6 +159,35 @@ def test_run_task_basic(self):
 
     self.assertEqual(task_manager.global_exit_code, 1)
 
+  @unittest.skipIf(sys.platform == 'win32', 'non-Windows specific')
```
pbos: Can you fake sys.platform so we can test it on all platforms?

volo-zyko: What do you mean? You want non-Windows test run on Windows?

pbos: Yes, exactly (through pretending that we run on Windows).

volo-zyko: The test will fail. save_file will end with .cache/gtest-parallel while test expects AppData/Local/gtest-parallel.

I don't see a point in running such test. Can you please explain why you need it because otherwise I don't understand how I can fix it? As for me having to 2 tests, 1 for Windows and 1 for anything else, should be enough.

pbos: I would expect you to mock out sys.platform (similar to os.environ above), so that we can run windows tests even when not on a Windows platform. I don't expect anyone to run these tests on Windows, so it's good to still have the same test coverage. Does that make sense?

volo-zyko: Ok, I see. But then mocking out just sys.platform is not enough. One has to mock os.path.expanduser too.

> I don't expect anyone to run these tests on Windows

I don't know. There are a lot of people work on Windows. As for my case we definitely will run gtes-parallel on Windows. So, Windows specific code will definitely run.

> Does that make sense?

This will definitely increase coverage but maintaining such tests will be more difficult. I'll see what I can do.

gtest_parallel_tests.py
```diff
@@ -159,35 +148,106 @@ def test_run_task_basic(self):
...
-  def test_get_save_file_path_unix(self):
-    self.assertEqual(os.path.join(os.path.expanduser('~'),
-                                  '.cache', 'gtest-parallel'),
-                     gtest_parallel.get_save_file_path())
 
-    with guard_environ('XDG_CACHE_HOME', os.getcwd()):
-      self.assertEqual(os.path.join(os.getcwd(), 'gtest-parallel'),
-                       gtest_parallel.get_save_file_path())
+@contextlib.contextmanager
+def guard_environ(var, val):
```
pbos: Is this something we can mock out as os.environ = new_environ, and finally: os.environ = old_environ? Instead of replacing values in the environment I'd rather we replaced it (so nothing in the tests can accidentally rely on the test environment).

pbos: Maybe name it test_environ, so the calls are with test_environ({'XDG_..': 'dir'}):

gtest_parallel_tests.py
```diff
@@ -159,35 +148,106 @@ def test_run_task_basic(self):
...
+  try:
+    temp_subdir = os.path.join(temp_dir, *path)
+    os.makedirs(temp_subdir)
+    yield temp_subdir
+  finally:
+    shutil.rmtree(os.path.join(temp_dir, path[0]))
+
+
+@contextlib.contextmanager
+def guard_patch_module(import_name, new_val):
```
pbos, line 174: Does it make sense to rename these test_patch_module? Is there a praxis to call them guard_?

pbos (commented):
Sorry for the slow feedback / being overall picky. Two things then I think we're done: Can we reasonably replace the environment instead of patching it + does it make sense to prefix guard_ functions with test_ instead?

gtest_parallel_tests.py
volo-zyko: 1) Ok, I used guard_patch_module to patch the whole os.environ.

2) I removed guard_environ as it's no longer needed.

gtest_parallel_tests.py
volo-zyko: All the guards implement contextmanager protocol. They don't test anything, they only guard that global changes (which can influence other tests) are correctly reverted. I can rename them but test_ is not a good candidate for this name either.

pbos:
Thanks!

PR #34, opened 2017-05-26
Add more tests to gtest-parallel.

ehlemur-zz:
We test TaskManager when we retry failed tests.

### DIFF
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -38,7 +38,7 @@ class LoggerMock(object):
                              self.exit_codes[test_id])
     test_lib.assertListEqual(expected['last_execution_time'][:retries],
                              self.last_execution_times[test_id])
-    test_lib.assertListEqual(expected['execution_number'],
+    test_lib.assertListEqual(expected['execution_number'][:retries],
                                   self.execution_numbers[test_id])
 
 
@@ -134,13 +134,13 @@ class TestTaskManager(unittest.TestCase):
         # Fails twice, then succeeds
         (('yet_another_binary', 'Fake.Test.FailTwice'), {
             'runtime_ms': [23, 25, 24],
-            'exit_code': [2, 2, 0],
+            'exit_code': [4, 4, 0],
             'last_execution_time': [None, None, 24],
         }),
         # Failing task
         (('fake_binary', 'Fake.FailingTest'), {
             'runtime_ms': [20, 30, 40],
-            'exit_code': [1, 1, 1],
+            'exit_code': [1, 2, 3],
             'last_execution_time': [None, None, None],
         })
     ]
@@ -150,19 +150,20 @@ class TestTaskManager(unittest.TestCase):
     self.times.assertRecorded(self,test_id, expected, retries)
     self.test_results.assertRecorded(self,test_id, expected, retries)
 
-  def test_run_task_basic(self):
+  def retry_failed_n_times(self, retry_failed, n_tasks, expected_exit_code):
     repeat = 1
-    retry_failed = 0
 
     task_mock_factory = TaskMockFactory(dict(self.test_data))
     task_manager = gtest_parallel.TaskManager(
         self.times, self.logger, self.test_results,
         task_mock_factory, retry_failed, repeat)
 
-    for test_id, expected in self.test_data:
+    for test_id, expected in self.test_data[:n_tasks]:
+      if expected_exit_code == 0:
+        print test_id, expected
       task = task_mock_factory.get_task(test_id)
       task_manager.run_task(task)
-      expected['execution_number'] = range(retry_failed + 1)
+      expected['execution_number'] = range(len(expected['exit_code']))
       self.assertTaskWasExecuted(test_id, expected, retry_failed + 1)
 
     self.assertEqual(len(task_manager.started), 0)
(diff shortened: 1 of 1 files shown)


ehlemur-zz:
@pbos @kwiberg 
PTAL :)

gtest_parallel_tests.py
```diff
@@ -150,19 +150,20 @@ def assertTaskWasExecuted(self, test_id, expected, retries):
     self.times.assertRecorded(self,test_id, expected, retries)
     self.test_results.assertRecorded(self,test_id, expected, retries)
 
-  def test_run_task_basic(self):
+  def retry_failed_n_times(self, retry_failed, n_tasks, expected_exit_code):
```
pbos: Can you make these named parameters, and below run with retry_failed=X instead of X, this makes the function calls easier to read.

gtest_parallel_tests.py
```diff
@@ -173,8 +174,19 @@ def test_run_task_basic(self):
         sorted(task.task_id for task in task_manager.failed),
         sorted(task.task_id for task in task_mock_factory.failed))
 
-    self.assertEqual(task_manager.global_exit_code, 1)
+    self.assertEqual(task_manager.global_exit_code, expected_exit_code)
+
+  def test_run_task_basic(self):
```
pbos: Can you rename all of these tests to what they actually verify? Such as "test_running_without_failures_should_do_X" or something?

pbos: This is because it's really not obvious what run_task_retry_failed_once tests vs. what run_task_retry_failed_twice tests.

pbos (changes requested):

ehlemur-zz:
PTAL

gtest_parallel_tests.py
```diff
@@ -114,56 +114,50 @@ def run(self):
...
             'exit_code': [0],
             'last_execution_time': [10],
-        }),
-        # Fails once, then succeeds
-        (('another_binary', 'Fake.Test.FailOnce'), {
+        })
+    self.failing_task = (
+        ('fake_binary', 'Fake.FailingTest'), {
+            'runtime_ms': [20, 30, 40],
+            'exit_code': [1, 2, 3],
```
pbos: I think you can just use 1 for failure here, there's no obvious thing that's verified by these codes being different.

gtest_parallel_tests.py
```diff
@@ -114,56 +114,50 @@ def run(self):
...
+            'last_execution_time': [None, None, None],
+        })
+    self.fails_once_then_succeeds = (
+        ('another_binary', 'Fake.Test.FailOnce'), {
             'runtime_ms': [21, 22],
             'exit_code': [3, 0],
```
pbos: Ditto: 1,0

gtest_parallel_tests.py
```diff
@@ -114,56 +114,50 @@ def run(self):
...
             'last_execution_time': [None, 22],
-        }),
-        # Fails twice, then succeeds
-        (('yet_another_binary', 'Fake.Test.FailTwice'), {
+        })
+    self.fails_twice_then_succeeds = (
+        ('yet_another_binary', 'Fake.Test.FailTwice'), {
             'runtime_ms': [23, 25, 24],
-            'exit_code': [2, 2, 0],
+            'exit_code': [4, 4, 0],
```
pbos: 1,1,0

gtest_parallel_tests.py
```diff
@@ -173,8 +167,40 @@ def test_run_task_basic(self):
...
+                       retries=2, expected_exit_code=3)
+
+    self.execute_tasks(tasks=[self.passing_task, self.failing_task],
+                       retries=2, expected_exit_code=3)
+
+  def test_task_that_fails_once_succeeds_with_two_attempts(self):
+    self.execute_tasks(tasks=[self.fails_once_then_succeeds],
+                       retries=1, expected_exit_code=0)
+
+  def test_task_that_fails_twice_fails_with_two_attempts(self):
```
pbos: _fails_with_one_retry(

pbos: Or: test_task_gets_repeated_only_once( if that's what you actually want to verify.

pbos: Or test_task_gets_repeated_according_to_retries(

Your call. Do add a comment that says # Executes test and retries once, not enough for the test to start passing, so gtest-parallel should return an error.

Add similar comments where you think they are necessary.

gtest_parallel_tests.py
```diff
@@ -173,8 +167,40 @@ def test_run_task_basic(self):
...
+    self.execute_tasks(tasks=[self.fails_twice_then_succeeds],
+                       retries=1, expected_exit_code=4)
 
+  def test_all_tasks_eventually_succeed(self):
```
pbos: test_repeats_tasks_until_they_succeed(
test_runner_succeeds_when_all_tasks_eventually_succeeds(

or something, since you verify that gtest-parallel returns success.

If you think this makes sense, otherwise push back to keep as is.

pbos (changes requested):
Thanks, this looks very good, some remaining comments.

ehlemur-zz:
PTAL 😄

pbos:
lgtm

PR #37, opened 2017-06-15
Add --run_test_cases command line option for running whole test cases in parallel

volo-zyko:
Sometimes it is safer to run whole test cases instead of individual tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,14 +59,10 @@ your friends here.
 
 ## Running tests cases in parallel
 
-Sometimes tests in test case make system-wide changes in their
-`SetUp()/SetUpTestCase()` methods. Examples of such changes are opening sockets
-on fixed address:port, writing to files at fixed location, etc. Running such
-code in parallel processes most likely will fail. In order to avoid this
-situation `gtest-parallel` provides `--serialize_test_cases` capability.
-
-    # ./gtest-parallel --serialize_test_cases path/to/binary
-
-This option can be used along with other `gtest-parallel` options and it ensures
-that tests from the same test case are never run in parallel. Naturally this
-works as long as other tests cases do not do the same globally visible things.
+Sometimes tests within a single test case use globally-shared resources (instead
+of temporary files or fake/mock objects) and cannot be run in parallel.
+Examples of such changes are opening sockets on fixed `address:port`, writing
+to files at fixed location, etc. Running such tests in parallel will either fail
+or be flaky. For test binaries where tests are not self contained, but test
+cases are, --serialize_test_cases can be provided to prevent them from running
+in parallel, while still benefiting from parallel speedup between test cases.
--- a/README.md
+++ b/README.md
@@ -56,3 +56,17 @@ Note that repeated tests do run concurrently with themselves for efficiency, and
 as such they have problem writing to hard-coded files, even if they are only
 used by that single test. `tmpfile()` and similar library functions are often
 your friends here.
+
+## Running tests cases in parallel
+
+Sometimes tests in test case make system-wide changes in their
+`SetUp()/SetUpTestCase()` methods. Examples of such changes are opening sockets
+on fixed address:port, writing to files at fixed location, etc. Running such
+code in parallel processes most likely will fail. In order to avoid this
+situation `gtest-parallel` provides `--serialize_test_cases` capability.
+
+    # ./gtest-parallel --serialize_test_cases path/to/binary
+
+This option can be used along with other `gtest-parallel` options and it ensures
+that tests from the same test case are never run in parallel. Naturally this
+works as long as other tests cases do not do the same globally visible things.
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -428,18 +428,27 @@ def find_tests(binaries, additional_args, options, times):
     for line in test_list.split('\n'):
       if not line.strip():
         continue
-      if line[0] != " ":
+      elif line[0] != " ":
         # Remove comments for typed tests and strip whitespace.
         test_group = line.split('#')[0].strip()
-        continue
-      # Remove comments for parameterized tests and strip whitespace.
-      line = line.split('#')[0].strip()
-      if not line:
-        continue
+        if not options.run_test_cases:
+          continue
 
-      test_name = test_group + line
-      if not options.gtest_also_run_disabled_tests and 'DISABLED_' in test_name:
-        continue
+        test_name = test_group + '*'
+        # Manually disable tests according to gtest_filter.
+        separator_index = options.gtest_filter.find('-')
+        if separator_index >= 0:
+          test_name += ':' + options.gtest_filter[separator_index:]
+      elif not options.run_test_cases:
+        # Remove comments for parameterized tests and strip whitespace.
+        line = line.split('#')[0].strip()
+        if not line:
+          continue
+
+        test_name = test_group + line
+        if (not options.gtest_also_run_disabled_tests and
+              'DISABLED_' in test_name):
+          continue
 
       last_execution_time = times.get_test_time(test_binary, test_name)
       if options.failed and last_execution_time is not None:
@@ -538,6 +547,9 @@ def main():
   parser.add_option('--timeout', type='int', default=None,
                     help='Interrupt all remaining processes after the given '
                          'time (in seconds).')
+  parser.add_option('--run_test_cases', action='store_true', default=False,
+                    help='run whole test cases in parallel '
+                         '(as opposed to individual tests)')
 
   (options, binaries) = parser.parse_args()
 
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -558,8 +558,8 @@ def main():
                     help='Interrupt all remaining processes after the given '
                          'time (in seconds).')
   parser.add_option('--serialize_test_cases', action='store_true',
-                    default=False, help='make sure that tests from the same '
-                                        'test case are never run in parallel')
+                    default=False, help='Do not run tests from the same test '
+                                        'case in parallel.')
 
   (options, binaries) = parser.parse_args()
 
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -428,27 +428,18 @@ def find_tests(binaries, additional_args, options, times):
     for line in test_list.split('\n'):
       if not line.strip():
         continue
-      elif line[0] != " ":
+      if line[0] != " ":
         # Remove comments for typed tests and strip whitespace.
         test_group = line.split('#')[0].strip()
-        if not options.run_test_cases:
-          continue
+        continue
+      # Remove comments for parameterized tests and strip whitespace.
+      line = line.split('#')[0].strip()
+      if not line:
+        continue
 
-        test_name = test_group + '*'
-        # Manually disable tests according to gtest_filter.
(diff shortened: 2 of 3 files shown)


pbos:
Does this fix an actual problem for you, or is this theoretical?

If I read this correctly this probably breaks existing DISABLED_ behavior, doesn't display individual tests running on the command line, nor list the correct number of tests run (I'm reading this as all tests in TestGroup becomes one test invocation, which might be wrong, and the DISABLED_ check is not part of this branch).

If this fixes an actual problem, and you still get significant speedup from gtest-parallel, I'd prefer if this was done so that whenever workers try to poll a new job, they try to find test groups that are not currently running, rather than splitting the tests differently when listing them.

On a found job, mark test_group_is_running["TestGroup"] = True (but probably use a set), and whenever a job finishes mark test_group_is_running["TestGroup"] = False (but probably remove from the proposed set).

If they fail to find one such job, there are fewer test groups remaining to run instead of workers, so joining the current thread is OK (act as if list was empty).

volo-zyko:
> Does this fix an actual problem for you, or is this theoretical?

Yes, it definitely fixes actual problem. We have test cases which do system wide things (opening sockets, creating files) in their `SetUp()`/`SetUpTestCase()` method. Running such tests in parallel predictably fails. However, running whole test cases in parallel (we have quite a lot of them in one executable - test runner) works fine.

> If I read this correctly this probably breaks existing DISABLED_ behavior

No, it doesn't. First of all, the old behavior (running tests in parallel) is not changed. When we run tests cases in parallel we execute test runner with option `--gtest_filter=TestGroup.*`. If there any disabled tests in `TestGroup` then those are handled by gtest at run time. Also, if gtest-parallel is run with --gtest_filter option itself then this is also handled by the new behavior. Below are different gtest-parallel invocations along with test runner's command line.

```sh
$> gtest-parallel --run_test_cases --gtest_filter=TestGroup.* TestRunner
    TestRunner --gtest_filter=TestGroup.*

$> gtest-parallel --run_test_cases --gtest_filter=TestGroup.* --gtest_also_run_disabled_tests TestRunner
    TestRunner --gtest_filter=TestGroup.* --gtest_also_run_disabled_tests

$> gtest-parallel --run_test_cases --gtest_filter=-*.HeavyTest TestRunner
    TestRunner --gtest_filter=TestGroup1.*:-*.HeavyTest
    TestRunner --gtest_filter=TestGroup2.*:-*.HeavyTest
    ....

# This will run all TestGroups with SomeTest which might be not what user expects
$> gtest-parallel --run_test_cases --gtest_filter=*.SomeTest TestRunner
    TestRunner --gtest_filter=TestGroup.*
```

However, while writing this reply I realized that --gtest_filter is a :-separated list and it can be much more complicated. So, this is not handled in proposed change... Anyway it can be fixed.

> doesn't display individual tests running on the command line

No, it doesn't but this is to be expected. Once user requests running test cases only test cases are displayed on the terminal. Certainly the full log with all running tests is available in file (gtest-parallel saves it for us).

> nor list the correct number of tests run

Again, this is to be expected. gtest-parallel shows progress in terms of executed test cases.

> I'd prefer if this was done so that whenever workers try to poll a new job, they try to find test groups that are not currently running, rather than splitting the tests differently when listing them.

Do I understand this correctly that we manually manage running tests and make sure that we never run tests from the same TestGroup in parallel?

pbos:
I'd like for this to be handled outside of the test listing, since we want `--retry-failed` to not trigger all of the test case again. A single task should still be a single test I think (or we create a lot of headache for us).

How about: You could refactor WorkerFn to optionally split `self.tasks` into test groups, and then replace the call to `task_manager.run_task` with a `for task in task_group: task_manager.run_task(task)`. I'd be happy with that, it sounds maintainable. :)

I'd like the command-line-option to state that it's not the normal case, so instead of saying `--run_test_cases` something like `--serialize_test_cases` makes more sense to me.

volo-zyko:
Well, I reverted my previous change and added new implementation as you initially suggested. I thought it was less intrusive.

Collecting tasks in task groups (your second suggestion) will heavily rearrange tests while tracking running test groups (your first suggestion) will better preserve original tasks ordering (if it was important). Later there can be a need for test shuffling. Currently test's ordering is not very random.

gtest_parallel.py
```diff
@@ -538,6 +557,9 @@ def main():
   parser.add_option('--timeout', type='int', default=None,
                     help='Interrupt all remaining processes after the given '
                          'time (in seconds).')
+  parser.add_option('--serialize_test_cases', action='store_true',
+                    default=False, help='make sure that tests from the same '
```
pbos: help string -> "Do not run tests from the same test case in parallel."

pbos (changes requested):
Can you figure out a way to test this too?

pbos:
Re test ordering being predictable right now, that's a bug. It should be ordered on last test runtime. I've a push for it here (but I want to add tests for this before pushing): https://github.com/google/gtest-parallel/pull/38

It still looks like you're going through the tasks in order though, so I think this shouldn't break the heuristic completely?

pbos:
Additionally, I added a README since last time. Can you put a subsection under flakiness for this feature and an example for where you might want to use it?

volo-zyko:
> so I think this shouldn't break the heuristic completely?

no, it shouldn't.

pbos:
@ehlemur can you help review if the test setup part of this?

gtest_parallel_tests.py
```diff
@@ -300,5 +303,133 @@ def test_get_save_file_path_win32(self):
...
+  class TaskManagerMock(object):
+    def __init__(self, group_checker):
+      self.group_checker = group_checker
+      self.check_lock = threading.Lock()
+
+    def run_task(self, task):
+      test_group = task.test_name.split('.')[0]
+
+      with self.check_lock:
+        self.group_checker.add(test_group)
```
ehlemur-zz: Why not get rid of the GroupChecker class and instead have a variable that stores whether tests from the same test groups have run in parallel.
Also, I think you should check all tasks were actually run.

gtest_parallel_tests.py
```diff
@@ -300,5 +303,133 @@ def test_get_save_file_path_win32(self):
...
+
+      # Delay as if real test were run.
+      time.sleep(0.001 * task.execution_number)
```
ehlemur-zz: Please don't use execution number for this. I find it a bit confusing.
Also, gtest_parallel.execute_tasks doesn't care what the tasks are, so you could have your own task class that stores only the test_name and a fake_execution_time.

gtest_parallel_tests.py
```diff
@@ -300,5 +303,133 @@ def test_get_save_file_path_win32(self):
...
+                                 task_manager, None, True)
+
+  def test_running_parallel_test_cases_with_repeats(self):
+    max_number_of_test_cases = 4
+    max_number_of_tests_per_test_case = 32
+    max_number_of_repeats = 16
+    max_number_of_workers = 16
+
+    tasks = []
+    for test_case in range(max_number_of_test_cases):
```
ehlemur-zz: Can you move the tests generation into its own function, please?

ehlemur-zz (commented):
Sure

gtest_parallel_tests.py
volo-zyko: I realized that I got meaning of task.execution_number wrong. However, even without this code tests should be fine.

gtest_parallel_tests.py
volo-zyko: done

gtest_parallel_tests.py
volo-zyko: done

ehlemur-zz:
Maybe you forgot to actually upload it?

gtest_parallel_tests.py
```diff
@@ -305,130 +305,96 @@ def test_get_save_file_path_win32(self):
...
-      def remove(self, test_group):
-        self.running_groups.remove(test_group)
+    tasks = self._generate_tasks(max_number_of_test_cases,
+                                 max_number_of_tests_per_test_case, 1)
+    expected_tasks_number = len(tasks)
 
-    task_manager = TestSerializeTestCases.TaskManagerMock(GroupChecker(self))
+    task_manager = TestSerializeTestCases.TaskManagerMock()
 
     gtest_parallel.execute_tasks(tasks, max_number_of_workers,
```
ehlemur-zz: Sorry, maybe you can also replace all these lines (347..358) with something like:
```
self._execute_tasks(max_number_of_test_cases=4, max_number_of_tests_per_test_case=32,
                    max_number_of_workers=16, serialize_test_cases=True)
```
instead of a `_generate_tasks()` function?
What do you think?

volo-zyko: Ok, you are right, there is code duplication.

ehlemur-zz:
Tests LGTM
@pbos do you agree?

README.md
```diff
@@ -56,3 +56,17 @@ Note that repeated tests do run concurrently with themselves for efficiency, and
 as such they have problem writing to hard-coded files, even if they are only
 used by that single test. `tmpfile()` and similar library functions are often
 your friends here.
+
+## Running tests cases in parallel
+
+Sometimes tests in test case make system-wide changes in their
+`SetUp()/SetUpTestCase()` methods. Examples of such changes are opening sockets
+on fixed address:port, writing to files at fixed location, etc. Running such
```
pbos: `address:port` with backticks.

README.md
```diff
@@ -56,3 +56,17 @@ Note that repeated tests do run concurrently with themselves for efficiency, and
...
+code in parallel processes most likely will fail. In order to avoid this
```
pbos: Running such tests in parallel will either fail or be flaky. For test binaries where tests are not self contained, but test cases are, `--serialize_test_cases` can be provided to prevent them from running in parallel, while still benefiting from parallel speedup between test cases.

README.md
```diff
@@ -56,3 +56,17 @@ Note that repeated tests do run concurrently with themselves for efficiency, and
 as such they have problem writing to hard-coded files, even if they are only
 used by that single test. `tmpfile()` and similar library functions are often
 your friends here.
+
+## Running tests cases in parallel
+
+Sometimes tests in test case make system-wide changes in their
```
pbos: Sometimes tests within a single test case use globally-shared resources (instead of temporary files or fake/mock objects) and cannot be run in parallel. Examples of...

pbos (commented):
Yeah, tests look good. Thanks!

README.md
```diff
@@ -56,3 +56,17 @@ Note that repeated tests do run concurrently with themselves for efficiency, and
...
+`SetUp()/SetUpTestCase()` methods. Examples of such changes are opening sockets
+on fixed address:port, writing to files at fixed location, etc. Running such
+code in parallel processes most likely will fail. In order to avoid this
+situation `gtest-parallel` provides `--serialize_test_cases` capability.
+
+    # ./gtest-parallel --serialize_test_cases path/to/binary
```
pbos: I think this part isn't necessary (command + below), the above paragraph should have enough info. :)

volo-zyko:
Well, I updated the README but, in general, with such massive changes you could have updated it yourself when integrating to upstream. I don't mind at all.

pbos:
Right, I'm less used to github command-line workflow. Will do, thanks. :)

PR #38, opened 2017-06-16
Sort test tasks based on last execution time.

pbos:
Permits tests to run faster, this functionality must have been lost in
some refactoring effort, and was not covered by tests.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -454,7 +454,7 @@ def find_tests(binaries, additional_args, options, times):
 
       test_count += 1
 
-  return tasks
+  return sorted(tasks)
 
 
 def execute_tasks(tasks, pool_size, task_manager, timeout):


pbos:
Can you help me suggest tests to add for this, so we make sure that we run tests in None -> long runtime -> short runtime order?

pbos:
Also interestingly, removing task_manager.run_task(task) from execution doesn't create any test failures. Eep. :)

ehlemur-zz:
lgtm
For testing it, maybe mock subprocess.Popen and test find_tests?
And yes, test coverage is quite poor right now hehe. I'll work on adding more tests.

gtest_parallel.py
```diff
@@ -454,7 +454,7 @@ def find_tests(binaries, additional_args, options, times):
 
       test_count += 1
 
-  return tasks
+  return sorted(tasks)
```
kwiberg, line 457: (Leaving comment here instead of on `Task.__lt__`, since Github won't let me add a comment on a line too far away from the change. Irritating...)

Looks good, except that Task implements `__lt__` but not the other comparison special functions, and the implementation of `__lt__` says that a < b _and_ b < a if a and b both have no last_execution_time. I'd suggest replacing it with
```
def __sorting_key(self):
  return (0 if self.last_execution_time is None else 1,
          self.last_execution_time,
          ...all the other fields here is some sane order...)

def __cmp__(self, other):
  return cmp(self.__sorting_key(), other.__sorting_key())
```
which solves both problems.

kwiberg (changes requested):

gtest_parallel.py
pbos: I'll do a follow-up change with this separately once I've added tests for and submitted this first. :)

volo-zyko: Sorry, for interfering but as far as I understand if `self.last_execution_time is None` then it should correspond to bigger value in a key. As for me the following is easier to understand:

```
def __sorting_key(self):
    return (sys.float_info.max if x.last_execution_time is None else x.last_execution_time,
            ...all the other fields here is some sane order...)
```

This makes sense as we want to run first those new/failed tests and tests with long execution time.

pbos:
Sorry I've been punting tests for this for way too long. I think it's more important to fix the regression now than wait for me to add the tests, since it makes the script run significantly faster. I've added a follow-up task.

PR #40, opened 2017-07-03
Race in TestTimes loading/saving

volo-zyko:
The race happens when there are several gtest-parallel instances are run in parallel.

For us it happens because we have a number of components each with own tests and we have targets in our build system which allow running tests for each component independently. Running all tests (for which we also have aggregate target) each time takes substantial amount of time and is not very convenient during active development. So, something like this: `make -j3 RunFooComponentTests RunBarComponentTests RunBazComponentTests` spawns for us 3 python processes of gtest-parallel running tests for FooComponent, BarComponent, and BazComponent respectively.

Now, how shall we fix it? As far as I understand the only portable and reliable way in python is a lock file. What do you think?

Also, currently suppressing/catching `struct.error` is acceptable for us but, in general, this is not how TestTimes is supposed to work.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -373,18 +373,16 @@ class TestTimes(object):
     def __enter__(self):
       self._fo = open(self._filename, self._mode)
 
+      # Regardless of opening mode we always seek to the beginning of file.
+      # This simplifies code working with LockedFile and also ensures that
+      # we lock (and unlock below) always the same region in file on win32.
+      self._fo.seek(0)
+
       try:
         if sys.platform == 'win32':
-          # We make sure that we lock (and unlock below) always
-          # the same region in file.
-          position = self._fo.tell()
-          self._fo.seek(0)
-
           # We are locking here fixed location in file to use it as
           # an exclusive lock on entire file.
           msvcrt.locking(self._fo.fileno(), msvcrt.LK_LOCK, 1)
-
-          self._fo.seek(position)
         else:
           fcntl.flock(self._fo.fileno(), fcntl.LOCK_EX)
       except IOError:
@@ -401,12 +399,8 @@ class TestTimes(object):
 
       try:
         if sys.platform == 'win32':
-          position = self._fo.tell()
           self._fo.seek(0)
-
           msvcrt.locking(self._fo.fileno(), msvcrt.LK_UNLCK, 1)
-
-          self._fo.seek(position)
         else:
           fcntl.flock(self._fo.fileno(), fcntl.LOCK_UN)
       finally:
@@ -464,6 +458,7 @@ class TestTimes(object):
         # We erase data from file while still holding a lock to it. This
         # way reading old test times and appending new ones are atomic
         # for external viewer.
+        fd.seek(0)
         fd.truncate()
         with gzip.GzipFile(fileobj=fd, mode='wb') as gzf:
           cPickle.dump(times, gzf, cPickle.HIGHEST_PROTOCOL)
@@ -472,10 +467,6 @@ class TestTimes(object):
 
   @staticmethod
   def __read_test_times_file(fd):
-    # We seek to the beginning of file because if it was opened with 'a' mode
-    # then the file's position is at the end of file.
-    fd.seek(0)
-
     try:
       with gzip.GzipFile(fileobj=fd, mode='rb') as gzf:
         times = cPickle.load(gzf)
@@ -485,9 +476,6 @@ class TestTimes(object):
       return None
     else:
       return times
-    finally:
-      # Seek to the beginning of file again to simplify working with it later.
-      fd.seek(0)
 
 
 def find_tests(binaries, additional_args, options, times):
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -30,6 +30,12 @@ import threading
 import time
 import zlib
 
+if sys.platform == 'win32':
+  import msvcrt
+else:
+  import fcntl
+
+
 # An object that catches SIGINT sent to the Python process and notices
 # if processes passed to wait() die by SIGINT (we need to look for
 # both of those cases, because pressing Ctrl+C can result in either
@@ -358,6 +364,37 @@ class CollectTestResults(object):
 
 # Record of test runtimes. Has built-in locking.
 class TestTimes(object):
+  class LockedFile(object):
+    def __init__(self, filename, mode):
+      self._filename = filename
+      self._mode = mode
+      self._fo = None
+
+    def __enter__(self):
+      self._fo = open(self._filename, self._mode)
+
+      try:
+        if sys.platform == 'win32':
+          msvcrt.locking(self._fo.fileno(), msvcrt.LK_LOCK, 1)
+        else:
+          fcntl.flock(self._fo.fileno(), fcntl.LOCK_EX)
+      except IOError:
+        self._fo.close()
+        raise
+
+      return self._fo
+
+    def __exit__(self, exc_type, exc_value, traceback):
+      try:
+        if sys.platform == 'win32':
+          msvcrt.locking(self._fo.fileno(), msvcrt.LK_UNLCK, 1)
+        else:
+          fcntl.flock(self._fo.fileno(), fcntl.LOCK_UN)
+      finally:
+        self._fo.close()
+
+      return exc_value is None
+
   def __init__(self, save_file):
     "Create new object seeded with saved test times from the given file."
     self.__times = {}  # (test binary, test name) -> runtime in ms
@@ -367,9 +404,10 @@ class TestTimes(object):
     self.__lock = threading.Lock()
 
     try:
-      with gzip.GzipFile(save_file, "rb") as f:
-        times = cPickle.load(f)
-    except (EOFError, IOError, cPickle.UnpicklingError, zlib.error):
+      with TestTimes.LockedFile(save_file, 'rb') as fd:
+        with gzip.GzipFile(fileobj=fd, mode='rb') as gzf:
+          times = cPickle.load(gzf)
+    except (EOFError, OSError, IOError, cPickle.UnpicklingError, zlib.error):
       # File doesn't exist, isn't readable, is malformed---whatever.
       # Just ignore it.
       return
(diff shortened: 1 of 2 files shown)


ehlemur-zz:
Just an idea. What about specifying a different test times file for each component?

pbos:
I think we should be able to run gtest-parallel in parallel for the same set of tests.

What about the following:

Pre-run: Try to open test times (reading), load if possible. If this fails because the file is busy, log a warning, but do nothing.
During run: Record test times only for the running test (do not store into existing structure).
After run: Try to open test time for reading and writing, read struct, update test times only for tests that were run. If this fails (open for rw), log a warning but do nothing.

So long as the tests are not starting while others are running, or not finishing at the exact same time, we do get test time updates properly from all runners even, and we never overwrite test times for tests that we haven't run.

volo-zyko:
@ehlemur Not clear how to choose file names for test times files and how to reuse them in subsequent runs. Probably this can be implemented but it will be complicated for sure.

@pbos I don't see how to implement what you propose given that all gtest-parallel work with one file and we can never guarantee that finishing one gtest-parallel will not coincide with starting another gtest-parallel (other people/projects might run gtest-parallel completely differently). Even if we are ready to sacrifice some of results of gtest-parallel it's not correct. It's just probably easier to implement, right?

volo-zyko:
The prototype is not perfect. In particular, file locking will not work on Unix if TestTimes is used from multiple threads (file locks are always per-process as they are associated with file descriptors) and file cannot be locked multiple times on Windows (see docs for msvcrt.locking). But as long as the usage is simple it works fine.

volo-zyko:
BTW, I forgot to mention that I tested the prototype on Windows, Linux and macOS.

pbos:
I'm out on vacation this week, I'll make sure to take a look next week. I'm worried about adding a lock file since that'll be stale if the process dies.

I don't think it's super important to reliably record all test times, since they will be eventually correct or correct most of the time. I think that's what we tried to do here. Maybe it's painful to miss reading the times, if so we could retry opening the file X times, sleeping 10ms in between.

When do we get a `struct.error`? Are multiple processes writing to the file? Shouldn't open(write) be an exclusive handle?

pbos:
Oh, I think you're describing what I just assumed open() did by default. Awesome^H^H^H^H.. Interesting that we can have multiple readers and writers at the same time. :(

Should we use os.open() to open the fd instead, since we can provide `O_EXCL` to it?

I'll make sure to take a look Monday.

volo-zyko:
> I'm worried about adding a lock file since that'll be stale if the process dies.

I used file locking mechanisms provided by operating system. It does not really create a lock file anywhere on the file system. Instead OS maintains some lock internally and this lock is associated with file handle which we use (handle to opened ~/.cache/gtest-parallel that is). Once process dies or abnormally exits all file handles are closed and locks are freed by OS.

> When do we get a `struct.error`?

As far as I can see when 1 process writes to file and another process reads from it at the same time.

> Are multiple processes writing to the file?

That's definitely possible in our case.

> Shouldn't open(write) be an exclusive handle?

This will not work. See what POSIX specification says about `O_EXCL`:

O_EXCL
    If O_CREAT and O_EXCL are set, open() shall fail if the file exists. The check for the existence of the file and the creation of the file if it does not exist shall be atomic with respect to other threads executing open() naming the same filename in the same directory with O_EXCL and O_CREAT set. If O_EXCL and O_CREAT are set, and path names a symbolic link, open() shall fail and set errno to [EEXIST], regardless of the contents of the symbolic link. If O_EXCL is set and O_CREAT is not set, the result is undefined.

So, this is primarily for creating files atomically. In our case the file will exist most of the time.

gtest_parallel.py
```diff
@@ -398,9 +436,20 @@ def record_test_time(self, binary, testname, runtime_ms):
   def write_to_file(self, save_file):
     "Write all the times to file."
     try:
-      with open(save_file, "wb") as f:
-        with gzip.GzipFile("", "wb", 9, f) as gzf:
-          cPickle.dump(self.__times, gzf, cPickle.HIGHEST_PROTOCOL)
+      with TestTimes.LockedFile(save_file, 'a+b') as fd:
+        fd.seek(0)
```
pbos: Is this necessary?

gtest_parallel.py
```diff
@@ -398,9 +436,20 @@ def record_test_time(self, binary, testname, runtime_ms):
...
+        try:
+          with gzip.GzipFile(fileobj=fd, mode='rb') as gzf:
+            times = cPickle.load(gzf)
+
+          times.update(self.__times)
+        except (EOFError, IOError, cPickle.UnpicklingError, zlib.error):
+          times = self.__times
+
+        fd.seek(0)
```
pbos: Can you put a comment here that we're erasing the file while we have the lock instead of reopening it so that we maintain exclusive access to it?

gtest_parallel.py
```diff
@@ -398,9 +436,20 @@ def record_test_time(self, binary, testname, runtime_ms):
...
-        with gzip.GzipFile("", "wb", 9, f) as gzf:
-          cPickle.dump(self.__times, gzf, cPickle.HIGHEST_PROTOCOL)
+      with TestTimes.LockedFile(save_file, 'a+b') as fd:
+        fd.seek(0)
+        try:
+          with gzip.GzipFile(fileobj=fd, mode='rb') as gzf:
+            times = cPickle.load(gzf)
+
+          times.update(self.__times)
+        except (EOFError, IOError, cPickle.UnpicklingError, zlib.error):
```
pbos: Comment that we could open the file here but failed to parse it (probably broken), so we replace it with the new data instead.

gtest_parallel.py
```diff
@@ -358,6 +364,37 @@ def dump_to_file_and_close(self):
 
 # Record of test runtimes. Has built-in locking.
 class TestTimes(object):
+  class LockedFile(object):
+    def __init__(self, filename, mode):
+      self._filename = filename
+      self._mode = mode
+      self._fo = None
+
+    def __enter__(self):
```
pbos, line 373: Can we put some kind of assert here that we're calling this from the main thread? I think we only read/write test times from the main thread, right?

pbos (commented):
Thanks, this sounds reasonable to me then.

O_EXLOCK and O_SHLOCK would've been cool, but they're not available here. :(

gtest_parallel.py
volo-zyko: Yes, it's necessary. When opening file for appending ('a' mode) file's position is at the end of file.

gtest_parallel.py
volo-zyko: Looking at what can be raised during unzipping/unpickling why not just catch either everything (with bare `except`) or catch generic `Exception`? Documentation is rather vague for both modules. In particular, gzip doesn't specify what can be raised and cPickle.load can raise a lot more than currently we have in `except`.

gtest_parallel_tests.py
```diff
@@ -376,16 +376,18 @@ def test_running_parallel_tests(self):
 
 class TestTestTimes(unittest.TestCase):
   def test_race_in_test_times_load_save(self):
-    max_number_of_workers = 16
-    max_numver_of_read_write_cycles = 128
+    max_number_of_workers = 1
```
volo-zyko: Sorry, I changed this for debugging and it looks like my testing was not complete. :(

gtest_parallel.py
volo-zyko: Currently test runs this code from multiple threads. So, this is not possible.

volo-zyko: I was wrong above in "file locking will not work on Unix if TestTimes is used from multiple threads". It will not work only if threads share the same file descriptor. Since file descriptors are global within process then all threads will see the file lock for given file descriptor either as acquired or released. In our case TestTimes hides actual file descriptor and consequently it will work as expected.

gtest_parallel.py
```diff
@@ -358,6 +364,42 @@ def dump_to_file_and_close(self):
...
+      self._fo = open(self._filename, self._mode)
+
+      try:
+        if sys.platform == 'win32':
+          msvcrt.locking(self._fo.fileno(), msvcrt.LK_LOCK, 1)
```
pbos: Can you put a comment here saying that we are locking a fixed location inside the file, to use as an exclusive lock on the entire file? Also, are we sure that `_fo` is always at position 0 here? Otherwise different `_fo` handles will try to lock different segments. If `self._mode` here is append, this is wrong.

pbos: Also, are any of these methods reentrant, or do they both fail if a thread locks a segment that another thread has locked?

volo-zyko: > Also, are we sure that _fo is always at position 0 here?

Good point. I'll add `self._fo.seek(0)` in `__enter__` and `__exit__` on win32.

> Also, are any of these methods reentrant, or do they both fail if a thread locks a segment that another thread has locked?

If I understood you correctly then both of them block until other thread releases the lock.

gtest_parallel.py
```diff
@@ -358,6 +364,56 @@ def dump_to_file_and_close(self):
...
+          # We make sure that we lock (and unlock below) always
+          # the same region in file.
+          position = self._fo.tell()
+          self._fo.seek(0)
+
+          # We are locking here fixed location in file to use it as
+          # an exclusive lock on entire file.
+          msvcrt.locking(self._fo.fileno(), msvcrt.LK_LOCK, 1)
+
+          self._fo.seek(position)
```
pbos: Do we need to support opening at end of file? Otherwise we can just always open at 0?

gtest_parallel.py
```diff
@@ -398,12 +453,42 @@ def record_test_time(self, binary, testname, runtime_ms):
...
+        with gzip.GzipFile(fileobj=fd, mode='wb') as gzf:
+          cPickle.dump(times, gzf, cPickle.HIGHEST_PROTOCOL)
     except IOError:
       pass  # ignore errors---saving the times isn't that important
 
+  @staticmethod
+  def __read_test_times_file(fd):
+    # We seek to the beginning of file because if it was opened with 'a' mode
+    # then the file's position is at the end of file.
+    fd.seek(0)
```
pbos: If we always open LockedFile at position 0 this shouldn't be needed either. WDYT?

pbos:
lgtm, thanks so much for making all of these changes.

PR #42, opened 2017-08-07
Add test which exhibits problem with log file paths on windows

volo-zyko:
The problem is that when we create log file the file path can be shorter than MAX_PATH in Windows (260 characters) but when we move it to 'passed'/'failed'/'interrupted' subdirectory the path can become longer and in such cases we miserably fail just before finishing gtest_parallel.

There are many options for fixing this issue but all of them have flaws. For example, for me it's ok to reduce log file name by removing test_binary from it.

### DIFF
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -414,6 +414,28 @@ class TestTestTimes(unittest.TestCase):
 
 
 class TestFilterFormat(unittest.TestCase):
+  def test_log_file_names(self):
+    self.assertEqual(
+      'bin-Test_case-100.log',
+      gtest_parallel.Task._logname('', 'bin', 'Test.case', 100))
+
+    self.assertEqual(
+      '../a/b/bin-Test_case_2-1.log',
+      gtest_parallel.Task._logname('../a/b', '../bin', 'Test.case/2', 1))
+
+    self.assertEqual(
+      '../a/b/bin-Test_case_2-5.log',
+      gtest_parallel.Task._logname('../a/b', '/c/d/bin', 'Test.case/2', 5))
+
+    self.assertEqual(
+      '/a/b/bin-Instantiation_Test_case_2-3.log',
+      gtest_parallel.Task._logname('/a/b', '../c/bin',
+                                   'Instantiation/Test.case/2', 3))
+
+    self.assertEqual(
+      '/a/b/bin-Test_case-1.log',
+      gtest_parallel.Task._logname('/a/b', '/c/d/bin', 'Test.case', 1))
+
   def test_on_windows_long_file_names_in_move_to(self):
     destination_subdir = 'interrupted'
     max_path = 260
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -18,6 +18,7 @@ import gtest_parallel
 import os.path
 import random
 import shutil
+import sys
 import tempfile
 import threading
 import time
@@ -410,6 +411,21 @@ class TestTestTimes(unittest.TestCase):
       finally:
         for worker in workers:
           worker.join()
+
+
+class TestFilterFormat(unittest.TestCase):
+  @unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")
+  def test_long_file_names_in_move_to(self):
+    class Task(object):
+      def __init__(self, length, dest):
+        self.log_file = os.path.join(dest, ('a' * length) + '.log')
+        with open(self.log_file, 'wb'):
+          pass
+
+    with guard_temp_dir() as temp_dir:
+      ff = gtest_parallel.FilterFormat(temp_dir)
+      tasks = [Task(l, temp_dir) for l in range(1, 255 - len(temp_dir))]
+      ff.move_to('passed', tasks)
 
 
 if __name__ == '__main__':
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -415,26 +415,36 @@ class TestTestTimes(unittest.TestCase):
 
 class TestFilterFormat(unittest.TestCase):
   def test_log_file_names(self):
+    def root():
+      return 'c:/' if sys.platform == 'win32' else '/'
+
     self.assertEqual(
       'bin-Test_case-100.log',
       gtest_parallel.Task._logname('', 'bin', 'Test.case', 100))
 
     self.assertEqual(
-      '../a/b/bin-Test_case_2-1.log',
-      gtest_parallel.Task._logname('../a/b', '../bin', 'Test.case/2', 1))
+      os.path.join('..', 'a', 'b', 'bin-Test_case_2-1.log'),
+      gtest_parallel.Task._logname(os.path.join('..', 'a', 'b'),
+                                   os.path.join('..', 'bin'),
+                                   'Test.case/2', 1))
 
     self.assertEqual(
-      '../a/b/bin-Test_case_2-5.log',
-      gtest_parallel.Task._logname('../a/b', '/c/d/bin', 'Test.case/2', 5))
+      os.path.join('..', 'a', 'b', 'bin-Test_case_2-5.log'),
+      gtest_parallel.Task._logname(os.path.join('..', 'a', 'b'),
+                                   os.path.join(root(), 'c', 'd', 'bin'),
+                                   'Test.case/2', 5))
 
     self.assertEqual(
-      '/a/b/bin-Instantiation_Test_case_2-3.log',
-      gtest_parallel.Task._logname('/a/b', '../c/bin',
+      os.path.join(root(), 'a', 'b', 'bin-Instantiation_Test_case_2-3.log'),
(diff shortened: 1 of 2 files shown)


pbos:
This sounds like a corner case where `log_dir=foo` works but not `=foo/interrupted`, or does windows truncate internally when this file is stored to `foo`? Can you give me the files/paths you're actually using when hitting this? Do you have ridiculous test name lengths?

Can we just spit out a warning when moving fails (e.g. do we get a good error message from Python here), or is the user "being weird" by having long paths to logdir or ridiculously long test names?

volo-zyko:
BTW, here is the output of the test:

```
test_long_file_names_in_move_to (gtest_parallel_tests.TestFilterFormat) ... ERROR

======================================================================
ERROR: test_long_file_names_in_move_to (gtest_parallel_tests.TestFilterFormat)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "gtest_parallel_tests.py", line 428, in test_long_file_names_in_move_to
    ff.move_to('passed', tasks)
  File "gtest_parallel.py", line 289, in move_to
    shutil.move(task.log_file, destination_dir)
  File "C:\Python27\lib\shutil.py", line 302, in move
    copy2(src, real_dst)
  File "C:\Python27\lib\shutil.py", line 130, in copy2
    copyfile(src, dst)
  File "C:\Python27\lib\shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] No such file or directory: 'c:\\users\\jenkin~1.sla\\appdata\\local\\temp\\tmptilqvr\\passed\\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.log'

----------------------------------------------------------------------
Ran 1 test in 1.188s

FAILED (errors=1)
```

Concerning real log files. For me normalized test_binary is a bigger problem. It includes full path to the test executable and if I build in deeply nested directory then this full path will be included in the log file name. Something like this: `_home_build_workspace_KH_Clang_Linux64_Release_Sanity_build_bin_TestBinary-CacheGrabberTestSuite_testGettingAllValuesSameCacheForDifferentCases-1.log`. This long file name is not a problem on Linux/macOS but it can easily exceed MAX_PATH on Windows.

volo-zyko:
In general we can use extended syntax `\\?\` in file names on Windows. In this case we can use 32k paths but this needs some caution. Not all Python utilities work with them. See this [article](https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx#maximum_path_length) about `\\?\` syntax.

pbos:
Agreed that _home.. (and that full path overall) is silly here. I'd be fine with replacing  `_home_build_workspace_KH_Clang_Linux64_Release_Sanity_build_bin_TestBinary` with `TestBinary`, e.g. using `basename`. Would that be sufficient (and ignoring `MAX_PATH` as an issue for now until we hit it again)?

volo-zyko:
Sorry for the delay.

Yes, for me using only `basename` in log file name is sufficient.

pbos:
So, only issue now is if we have `path/out/Release/test` and `path/out/Debug/test` running in the same execution (because they will have the same execution number and hence the same log file). This is a bit of an edge case but I can imagine people running the same tests in both debug and release configurations, and this could cause a lot of headache (the bug would be nasty to try to find).

Could we try to generate the smallest prefix (if basename(test) is unique for every test, use it, otherwise use deepest folder + basename(test) if they are unique, otherwise use the name down to the second deepest folder)?

I would also be fine if we simplify this check to `if (all basenames are unique) use basenames else use full path`. Do you think that check can fit somewhere?

pbos:
I think a good test to add is that `out/Release/test` and `out/Debug/test` should generate different logname files.

volo-zyko:
> if basename(test) is unique for every test

At the time when gtest_parallel runs, it has no knowledge about other instances of gtest_parallel running in parallel. So, I don't understand how we can check for uniqueness of test runners base names.

What we can do it is to generate a hash sum of dirname of test_binary and use it in log file names. This is easy to implement but will make file names a bit cryptic.

pbos:
That's fine, I was thinking when running `out/Release/test` and `out/Debug/test` from the same invocation of `gtest-parallel`.

E.g. `$ ./gtest-parallel out/{Release,Debug}/test`

volo-zyko:
I meant that the hash sum is calculated from `out/{Release,Debug}` part and added to the log file name but now I probably have a better idea.

We can calculate relative path from output directory to test binary and normalize it (probably with stripping `../` parts). I.e. for `$ ./gtest-parallel --output_dir logs out/{Release,Debug}/test` relative paths will be `../out/Release/test` and `../out/Debug/test`, after normalization: `out_Release_test`, `out_Debug_test`. So, once output directory and test binary are not very far from each other log file names will be short.

pbos:
How about we generate:

`['..', 'out', 'Release', 'test']`
`['..', 'out', 'Debug', 'test']`

for each path, and then find the smallest suffix ->

`['Release', 'test']`
`['Debug', 'test']`

and then normalize and put together:

`'Release_test'`
`'Debug_test'`

For normal invocations it'd very commonly yield:

`'test'`
`'test2'`

Problem is that we'd need to look at all tests together, but maybe we can go through all args initially and  generate a test -> test_log_name mapping before running them.

pbos:
I think I can take over this one if you agree on the solution above.

We might also need to consider invocations where people accidentally supply the binary name twice (since that yields conflicting log names), but I think we can error out on that rather than trying to append something.

volo-zyko:
> We might also need to consider invocations where people accidentally supply the binary name twice

This is already checked in `main()`.

```
  unique_binaries = set(os.path.basename(binary) for binary in binaries)
  assert len(unique_binaries) == len(binaries), (
      "All test binaries must have an unique basename.")
```

Concerning your solution. You assume that there is only one `gtest-parallel` running many binaries but it can be vice versa - many `gtest-parallel` each with one binary. In the latter scenario, the smallest suffix will be a binary name. In general, this is acceptable for me and I don't mind if you take it over.

pbos:
I think a good answer to many runners that have the same binary is to override `--output_dir`. We could also make sure that this is generated by default (say as `/tmp/gtest-parallel-randomhex`). Currently we remove the `--output_dir` content on startup with: `shutil.rmtree(options.output_dir)`, so logging here from two processes is racy with current behavior.

pbos:
Creating the randomhex directory would however mean that users have to clean them up themselves, unless we empty the temp directory by default (unless the user supplies --output_dir). I think the unique basename thing is more urgent though.

pbos:
Sorry I missed that we enforce unique basenames already. I think my unique prefix is overkill then and we can go with your solution for now (until someone needs `{Debug,Release}` to work).

Can you maybe break out the log file name function and test that it directly has the desired behavior instead of the windows-specific test that you currently have in there?

You can test that e.g.: 
`Test._logname(os.path.join(["..", "foo", "bar"]), "TestAThing/2", 2)` equals "bar-TestAThing_2-2.log" for instance?

Add some others for good measure that you can think of, maybe highlighting parameterized test cases or other that might be missed if we think tests are named like "TestAThing" and not "TestAThing/5".

pbos:
Does it sound good to you that we log with basename only and add documentation in the README on using `--output_dir` if you're using multiple parallel gtest-parallel invocations?

volo-zyko:
Sounds reasonable but this doesn't solve the problem with `{Release,Debug}/test`. If this is ok for you then let's do this. I add the tests and you update the README. :)

pbos:
Perfect, thanks! I don't need us to solve the `{Release,Debug}/test` problem now since it's not a regression, I was more worried about introducing one because of it. :)

gtest_parallel_tests.py
```diff
@@ -414,6 +414,28 @@ def test_times_worker():
 
 
 class TestFilterFormat(unittest.TestCase):
+  def test_log_file_names(self):
+    self.assertEqual(
+      'bin-Test_case-100.log',
+      gtest_parallel.Task._logname('', 'bin', 'Test.case', 100))
+
+    self.assertEqual(
+      '../a/b/bin-Test_case_2-1.log',
```
pbos: Should these be using `os.path.join(['..','a','b','bin...log'])` to get platform-dependent paths? Same for the input names? I'm guessing these tests might fail on Windows now which would use `..\a\b\bin-Test..log`?

volo-zyko: Oh, sorry. My fault. Now it definitely works on Windows.

pbos:
Looks good to me. I'll merge this but remove `test_long_file_names_in_move_to` which I think adds no value. I'll also change `c:/` to `C:\\`.

PR #43, opened 2017-08-07
converted project to a package

nicmcd:
I've converted gtest-parallel into a proper package format. This way users can more easily incorporate the project into their building and testing flow. This uses the standard package installation mechanism of setuptools. I've edited the README.md with a one-liner describing how to install.


pbos:
This breaks existing setups (buildbots etc. that expect `./gtest-parallel` to be present, and my personal `.bashrc`) and greatly complicates repository structure. Is it possible to add a `setuptools` repository that depends on this repository outside of this repository and pull in files as part of the installation step? I'd happily have a link to that repository in `README.md`, but I'd prefer to keep this repository simpler.

@kwiberg / @ehlemur in case they have different opinions as I've never used `setuptools` and don't know what the fuss is.

nicmcd:
The main reason to use packaging and setuptools is so that you could submit it to pypi and people could install it with `pip install gtest-parallel`. Also, the python standard is to locate binaries in `~/.local/bin` and libraries in `~/.local/lib`. Many people put `~/.local/bin` in their PATH in their `.bashrc`. This way `gtest-parallel` looks like any other executable on the machine.

I understand your concerns and see that it will certainly break legacy environments, however, I think the packaging will make it easier to adopt. Presumably these buildbots etc. all need a custom path to the gtest-parallel directory to know where to find the run script.

"Is it possible to add a setuptools repository that depends on this repository outside of this repository and pull in files as part of the installation step?"
Certainly. If you guys decide not to use packaging, I'll create such a repo, but it won't be pypi compatible, only manual install (`git clone ....`).

pbos:
Can you make the setup script copy and create the paths / files / directories in a temp directory and then cd and install from there? That way you won't need any `.gitignore` changes nor changing this repository, while still being able to use `pip install`?

E.g. set up `/tmp/fooXVHK` or whatever, `shutil.copy` files there under any structure you want in there, then `cd /tmp/fooXVHK` before running `setup(`. Does that make sense? Does it still support your setup workflow through `pip install gtest-parallel`?

I don't mind hosting the setup script here, but it would be nice to not have to the increased repository complexity for it.

Also, is there any sensible way to test this packetization, so that it's not breaking from reorganizing files?

nicmcd:
I'll work on a setup script that doesn't change directory layout. I'll submit a different pull request when complete.

pbos:
Thanks!

PR #45, opened 2017-08-28
Document --output_dir for multiple invocations.

pbos:
This is necessary to avoid a race between the two processes that would
otherwise try to delete each other's logs.


pbos:
@volo-zyko can you take a look as well?

kwiberg (approved):

volo-zyko:
Looks good to me.

PR #46, opened 2017-09-05
Log to gtest-parallel-logs/ subdir.

### LINKED ISSUE #44: Add a mechanism for avoiding removing non-gtest-parallel data in output_dir
Currently, if a user sets `--output_dir=/path/to/important/data`, `gtest-parallel` will clear that out from any files to remove files from old test runs.

Some different solutions would be:
1: Warn if the directory exists, but does not contain a `.gtest-parallel-logs` file (that we would start saving). This would be a bad upgrade path for existing directories, since it would require user intervention.
2: Log to `gtest-parallel/` subdirectory, and remove logs from in here. This would change the `--output_dir` default to `/tmp` instead of `/tmp/gtest-parallel`, which would then be joined into `/tmp/gtest-parallel` in the end. This heuristic would only fail if a user does `--output_dir=path/to/git-dir` when it contains gtest-parallel, which I think we can ignore.
3: We could also call this subdirectory in 2 `gtest-parallel-logs`, which should never overlap.
4: Only remove `*.log`, but eew.
5: Refuse to log to any directories that contain non-`*.log` files. Also eew.

I don't like 1, because if you run it once to /path/to/important/data, select "ok", and then run it again, we nuke important data anyways. I think I like 2 or 3 best, not sure which. Both these solutions require that users know that we log to a subdirectory. Though given that we do `passed/`, `failed/` and `interrupted` subdirectories already this might be OK, but it depends on if users are scripting against known subdirectories. @ehlemur, do you make use of how subdirectories under `--output_dir` are organized?

pbos:
This changes the default logdir from /tmp/gtest-parallel to
/tmp/gtest-parallel-logs. The primary motivation is that that we do this by
appending gtest-parallel-logs (to /tmp by default), so that if a user provides a
directory that is already in use, we wouldn't clear that directory out
completely (--output_dir=Docs/ was previously dangerous, as we'd nuke Docs/*).

Now instead we'd only clear out a subdirectory that we create and own ourselves.

Fixes #44.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -603,8 +603,11 @@ def main():
       usage = 'usage: %prog [options] binary [binary ...] -- [additional args]')
 
   parser.add_option('-d', '--output_dir', type='string',
-                    default=os.path.join(tempfile.gettempdir(), "gtest-parallel"),
-                    help='output directory for test logs')
+                    default=tempfile.gettempdir(),
+                    help='Output directory for test logs. Logs will be '
+                         'available under gtest-parallel-logs/, so '
+                         '--output_dir=/tmp will results in all logs being '
+                         'available under /tmp/gtest-parallel-logs/.')
   parser.add_option('-r', '--repeat', type='int', default=1,
                     help='Number of times to execute all the tests.')
   parser.add_option('--retry_failed', type='int', default=0,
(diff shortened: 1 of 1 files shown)


ehlemur-zz (approved):
LGTM :)

Issue #47, opened 2017-09-05
Add tests to make sure that tests run in proper order (failed or unseen tests first, longest running tests first)


pbos:
Follow-up task to #38. This path should be tested.

pbos:
Fixed by @ehlemur with #49.

PR #48, opened 2017-09-05
Replace __lt__ instance with sorting key.

pbos:
Missing test time (unseen tests or failing runs) still take precedence
over finished tests. This format is easier to read and removes the issue
where we only supplied one of the rich comparison methods. Now we supply
none.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -169,12 +169,16 @@ class Task(object):
     self.log_file = Task._logname(self.output_dir, self.test_binary,
                                   test_name, self.execution_number)
 
-  def __lt__(self, other):
-    if self.last_execution_time is None:
-      return True
-    if other.last_execution_time is None:
-      return False
-    return self.last_execution_time > other.last_execution_time
+  def __sorting_key(self):
+    # Unseen or failing tests (both missing execution time) take precedence over
+    # execution time.
+    return (1 if self.last_execution_time is None else 0,
+            self.last_execution_time)
+
+  def __cmp__(self, other):
+    # Reverse sorting order, we want larger items (longer runtimes or
+    # unseen/failing tests) first to run longer tests first.
+    return -cmp(self.__sorting_key(), other.__sorting_key())
 
   @staticmethod
   def _normalize(string):
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -171,14 +171,13 @@ class Task(object):
 
   def __sorting_key(self):
     # Unseen or failing tests (both missing execution time) take precedence over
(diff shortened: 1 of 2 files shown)


pbos:
Follow-up as promised, this should possibly/probably not be submitted without a fix for #47 (which would test this path too).

gtest_parallel.py
```diff
@@ -170,12 +170,16 @@ def __init__(self, test_binary, test_name, test_command, execution_number,
...
-    if self.last_execution_time is None:
-      return True
-    if other.last_execution_time is None:
-      return False
-    return self.last_execution_time > other.last_execution_time
+  def __sorting_key(self):
+    # Unseen or failing tests (both missing execution time) take precedence over
+    # execution time.
+    return (1 if self.last_execution_time is None else 0,
+            self.last_execution_time)
```
kwiberg: It may be best to put the minus sign here instead of below. Otherwise, you have to explain here why 1 gets sorted before 0.

pbos: since sorted(foo) is normally in increasing order, should we instead change the sorted call to `sorted(foo, reverse=True)`?

ehlemur-zz: That's clearer in my opinion.

pbos: Done, I'll wait until your tests are in to submit this regardless.

pbos:
PTAL, I updated the tests to match the new sorting order (tests' sorting keys are in ascending order as expected, but we expect them to be sorted in the reverse order, so they run longest tests first).

gtest_parallel_tests.py
```diff
@@ -393,10 +393,10 @@ def _call_find_tests(self, test_data, options=None):
         test_data.keys(), [], options, TestTimesMock(self, test_data))
     return tasks, popen_mock
 
-  def test_tasks_are_sorted(self):
+  def test_tasks_are_sorted_in_descending_order(self):
     tasks, _ = self._call_find_tests(
         self.MULTIPLE_BINARIES_MULTIPLE_TESTS_ONE_FAILURE)
-    self.assertEqual(tasks, sorted(tasks))
+    self.assertEqual(tasks, sorted(tasks, reverse=True))
```
ehlemur-zz: Now that I think about it, we should also do something like:
```
self.assertEqual([task.last_execution_time for task in tasks],
                 [None, 4, 4, 3, 2])
```
to make sure tasks are actually sorted as we want them to.

pbos: Agreed, PTAL.

ehlemur-zz (approved):

PR #49, opened 2017-09-11
Add (some) tests for gtest_parallel.find_tests

ehlemur-zz:


gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+      tasks = gtest_parallel.find_tests(
+          fake_tests.keys(), [],
+          self.MockOptions(gtest_also_run_disabled_tests=False),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertEqual(len(tasks), 1)
+    self.assertFalse("DISABLED_" in tasks[0].test_name)
+    self.assertNotIn("--gtest_also_run_disabled_tests", mock_Popen.last_invocation)
+
+    with guard_patch_module('subprocess.Popen', mock_Popen):
```
pbos: This looks like it's a separate test from above? You can have "does not run disabled tests when disabled" as a test separate from runs disabled tests when asked

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
                                    os.path.join(root(), 'c', 'd', 'bin'),
                                    'Test.case', 1))
 
+
+class TestFindTests(unittest.TestCase):
+  class MockOptions(object):
```
pbos: Should these be under this test case? I'd expect mock options to be something generic that we can reuse?

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+      self.gtest_also_run_disabled_tests = gtest_also_run_disabled_tests
+      self.gtest_color = gtest_color
+      self.gtest_filter = gtest_filter
+      self.output_dir = output_dir
+      self.repeat = repeat
+      self.shard_count = shard_count
+      self.shard_index = shard_index
+
+  class MockTestTimes(object):
+    def __init__(self, test_data=None):
```
pbos: Can this be `test_data={}` or will that refer to the same dictionary across invocations?

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+      self.failed = failed
+      self.gtest_also_run_disabled_tests = gtest_also_run_disabled_tests
+      self.gtest_color = gtest_color
+      self.gtest_filter = gtest_filter
+      self.output_dir = output_dir
+      self.repeat = repeat
+      self.shard_count = shard_count
+      self.shard_index = shard_index
+
+  class MockTestTimes(object):
```
pbos: Same, should shit really be under TestFindTests?

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+    def __init__(self, test_data=None):
+      self._test_data = test_data or {}
+
+    def get_test_time(self, test_binary, test_name):
+      test_group, test = test_name.split('.')
+      return self._test_data.get(test_binary, {}).get(test_group, {}).get(test, None)
+
+  class MockPopen(object):
```
pbos: Ditto

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+    self.assertEqual(len(tasks), 2)
+    self.assertEqual(sorted([task.test_name for task in tasks]),
+                     ["FakeTest.DISABLED_Test2", "FakeTest.Test1"])
+    self.assertIn("--gtest_also_run_disabled_tests", mock_Popen.last_invocation)
+
+  def test_runs_only_failed_tests_when_asked(self):
+    fake_tests = {
+        "fake_unittests": {
+            "FakeTest": {
+                "FailedTest": None,
```
pbos: Comment that failed tests have no recorded runtime (just like new tests).

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+          self.MockOptions(failed=False),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertEqual(len(tasks), 2)
+    self.assertEqual(sorted([task.test_name for task in tasks]),
+                     ["FakeTest.FailedTest", "FakeTest.Test"])
+    self.assertEqual(sorted([task.last_execution_time for task in tasks]),
+                     [None, 1])
+
+    with guard_patch_module('subprocess.Popen', mock_Popen):
```
pbos: This looks like two tests baked into one.

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+          self.MockOptions(gtest_also_run_disabled_tests=True),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertEqual(len(tasks), 2)
+    self.assertEqual(sorted([task.test_name for task in tasks]),
+                     ["FakeTest.DISABLED_Test2", "FakeTest.Test1"])
+    self.assertIn("--gtest_also_run_disabled_tests", mock_Popen.last_invocation)
+
+  def test_runs_only_failed_tests_when_asked(self):
+    fake_tests = {
```
pbos: Can these be shared as data, kOneFailedTestOnePassing or something. But proper style.

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+          fake_tests.keys(), [],
+          self.MockOptions(failed=True),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertEqual(len(tasks), 1)
+    self.assertEqual(tasks[0].test_binary, "fake_unittests")
+    self.assertEqual(tasks[0].test_name, "FakeTest.FailedTest")
+    self.assertIsNone(tasks[0].last_execution_time)
+
+  def test_processes_gtest_filter(self):
```
pbos: Should this be called something like test_find_tests_applies_gtest_filter ?

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+
+    with guard_patch_module('subprocess.Popen', mock_Popen):
+      gtest_parallel.find_tests(
+          fake_tests.keys(), [],
+          self.MockOptions(gtest_filter=""),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertNotIn('--gtest_filter=SomeFilter', mock_Popen.last_invocation)
+
+  def test_processes_gtest_color(self):
```
pbos: Same, phrase this as what it's testing

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+      )
+    self.assertIn('--gtest_filter=SomeFilter', mock_Popen.last_invocation)
+
+    with guard_patch_module('subprocess.Popen', mock_Popen):
+      gtest_parallel.find_tests(
+          fake_tests.keys(), [],
+          self.MockOptions(gtest_filter=""),
+          self.MockTestTimes(fake_tests)
+      )
+    self.assertNotIn('--gtest_filter=SomeFilter', mock_Popen.last_invocation)
```
pbos: Can this be foreach, assert that gtest_filter is gone (not just SomeFilter)

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+    fake_tests = {
+        "fake_unittests": {
+            "FakeTest": {
+                "TestSomething": None,
+            }
+        }
+    }
+    mock_Popen = self.MockPopen(fake_tests)
+
+    with guard_patch_module('subprocess.Popen', mock_Popen):
```
pbos: There's a lot of code repetition here (I've definitely seen this block more than once). Can you have a `def mock_find_tests(mock_popen, fake_tests, options):` method and remove all corresponding `with` blocks with one-liners?

gtest_parallel_tests.py
```diff
@@ -446,5 +446,207 @@ def root():
...
+
+  def test_processes_repeat(self):
+    fake_tests = {
+        "fake_unittests": {
+            "FakeTest": {
+                "TestSomething": None,
+            }
+        }
+    }
+    mock_Popen = self.MockPopen(fake_tests)
```
pbos: mock_popen, here and everywhere else (no uppercase P, right?)

gtest_parallel_mocks.py
```diff
@@ -48,14 +48,20 @@ def assertRecorded(self, test_id, expected, retries):
 
 
 class TestTimesMock(object):
-  def __init__(self, test_lib):
+  def __init__(self, test_lib, test_data=None):
```
pbos, line 51: Should this default to `{}` or will that not generate a new dictionary per usage?

gtest_parallel_mocks.py
```diff
@@ -141,3 +147,41 @@ def __init__(self, test_id, execution_number, test_data):
 
   def run(self):
     pass
+
+
+class OptionsMock(object):
+  def __init__(self, failed=False, gtest_also_run_disabled_tests=False,
```
pbos: Is there any way that we can assert that these options are the same as the default `optparse` options? I don't want us to maintain two copies of the default values without enforcing equality.

Maybe the optparse arguments could be broken out of main?

gtest_parallel_tests.py
```diff
@@ -335,5 +337,135 @@ def root():
                                    'Test.case', 1))
 
 
+class TestFindTests(unittest.TestCase):
+  ONE_DISABLED_ONE_PASSED_TEST = {
```
pbos: ONE_ENABLED ? Test1 hasn't passed (says None)

gtest_parallel_tests.py
```diff
@@ -335,5 +337,135 @@ def root():
...
+      }
+  }
+  ONE_TEST = {
+      "fake_unittests": {
+          "FakeTest": {
+              "TestSomething": None,
+          }
+      }
+  }
+  COMPLEX_TEST = {
```
pbos: Can you put something in here that describes what you want the test to contain (rather than just being complex), such as MULTIPLE_BINARIES_MULTIPLE_TESTS_ONE_FAILURE or whatever? Even if the name gets sucky-long, it's better than a non-descript name.

gtest_parallel_mocks.py
ehlemur-zz: It will re-use the same dictionary over multiple calls.

ehlemur-zz:
Ok, PTAL

gtest_parallel_mocks.py
pbos: Ew, nice foot shooter, Python.

PR #50, opened 2017-09-12
Move mocks for gtest_paralell_tests.py to another file.

ehlemur-zz:
What do you think?
I think it might make things easier to read.

### DIFF
--- a/gtest_parallel_mocks.py
+++ b/gtest_parallel_mocks.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python2
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


gtest_parallel_mocks.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python2
```
pbos: This doesn't need to be executable so you can remove this line, right? (probably applies for all but `gtest-parallel`)

gtest_parallel_mocks.py
```diff
@@ -0,0 +1,184 @@
...
+    self.test_binary = test_id[0]
+    self.test_name = test_id[1]
+    self.task_id = (test_id[0], test_id[1], execution_number)
+
+  def run(self):
+    pass
+
+
+@contextlib.contextmanager
+def guard_temp_dir():
```
pbos: This is not a mock. Either move this and below to another file (back to _tests) or rename this file something like `gtest_parallel_test_utils.py`. WDYT?

pbos (commented):
Looks good overall.

gtest_parallel_mocks.py
pbos: I think I prefer moving these and below back to the top of the test file (gtest_parallel_mocks.py is a good name). Up to you though.

ehlemur-zz:
PTAL :)

PR #51, opened 2017-09-21
Improve error reporting when test fails or crashes

volo-zyko:
Currently, when test finishes with non-zero exit code it's not possible to guess from logs what went wrong. Exit status gives a hint about it. For example, when a process gets a signal exit code is negative.


volo-zyko:
It turns out this functionality is already implemented. I close this pull request.

PR #52, opened 2017-10-13
Add --print_test_summary option.

pbos:
Useful for flakiness stats, prints X / Y passed stats at end of run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,23 @@ as such they have problem writing to hard-coded files, even if they are only
 used by that single test. `tmpfile()` and similar library functions are often
 your friends here.
 
+### Flakiness Summaries
+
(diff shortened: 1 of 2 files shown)


kwiberg (approved):
Looks good!

Would it make sense to enable this by default when the number of repetitions is >1?

pbos:
Good idea, I think we don't need the flag if we gate it on repeat >1 + any test has failed (I don't think we should print it when all tests are passing, since that adds noise). This also makes the feature way more discoverable. WDYT?

ehlemur-zz:
I like that.

pbos:
I'll merge this then, @kwiberg let me know if you have something post-review that you'd like me to fix.

PR #53, opened 2017-10-19
Specify python2.7 in shebang as this is required

alexweej:
Any system where python2 is actually python2.6 will fail to run this.

### DIFF
--- a/gtest-parallel
+++ b/gtest-parallel
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python2.7
 # Copyright 2017 Google Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");


pbos:
Do you have any idea of the prevalence of `python2.7` as a system binary (where `python2` is already v2.7)? I think this is the right thing to do if it doesn't break existing installations.

pbos:
To sample, @ehlemur @kwiberg can you check if `env python2.7` are available on your machines?

alexweej:
It's the default CPython configuration to have pythonX and pythonX.Y, only in very rare custom installations would someone have a python2 but not a 2.7. In our case we had python2->python2.6 and python2.7 due to... reasons... but this script definitely requires 2.7.

Hope that helps!

pbos:
Definitely does, thanks. My random Ubuntu 16.04.2 LTS machine has `python2.7` as well, as does the `python` package on Debian Testing. Do you know if this applies to OS X machines as well? I don't know the current state of the world on Macs at all.

If you'd prefer not to sign the CLA I can address this as well. Let me know.

volo-zyko:
> Do you know if this applies to OS X machines as well? I don't know the current state of the world on Macs at all.

macOS has python2.7 since OS X 10.9 (i.e. since 2013)

pbos:
Hi, I can't accept this patch as the CLA isn't signed. Are you able to sign it?

PR #55, opened 2017-10-30
Handle errors better when obtaining tests list.

volo-zyko:
If test runner is broken and crashes even before it handles `--gtest_list_tests` then gtest-parallel should probably fail too. Currently, it says `[0/0] Running tests...` and returns success status.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -526,9 +526,9 @@ def find_tests(binaries, additional_args, options, times):
       list_command += ['--gtest_filter=' + options.gtest_filter]
 
     try:
-      test_list = subprocess.Popen(list_command,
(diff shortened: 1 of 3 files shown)


pbos:
I think this looks good, do you think we could add a test for it somewhat easily?

volo-zyko:
There is a lot has changed since I last looked at this. Now, gtest-parallel has much more tests and I broke a number of them with my first change.

pbos:
@ehlemur can you help review this? :)

ehlemur-zz:
Sure :)
I'll check tomorrow

pbos:
Ping @ehlemur let me know if you need me to take this over.

ehlemur-zz (approved):
Sorry, I thought I had already reviewed. This looks good to me.

PR #59, opened 2017-12-09
(Fix issue #57) Disable by default log files saving

### LINKED ISSUE #57: Log file names generated by gtest-parallel are too long
I've been hit by this at least 2 times.

1) Evidently, on Windows, there is a restriction of MAX_PATH (260 characters) FS path length.

2) On encrypted partition in Ubuntu (using eCryptfs). See this [comment](https://unix.stackexchange.com/a/32834). Since eCryptfs protects not only file's content but also file names then there is an approximate restriction of 140 characters file name length.

There probably other cases when gtest-parallel can fail writing log file due to file/path length restrictions.

Now I'm considering options for fixing these issues all at once:

1) Make writing log files configurable and if it's disabled then write everything to console. Drawback: this will clutter the output.

2) Restrict file/path length. It seems we discussed this previously. For example, use hash of test name as log file name. Drawback: difficult to find log files for each specific test.

3) Try to use system extensions for writing longer file names. While this can be done on Windows it will not work for eCryptfs or any other FS without such extensions.

4) Propose yours.

I'm in favor of option 1. Which option is acceptable for gtest-parallel?

volo-zyko:
If --output_dir command line options is omitted then
gtest-parallel does not save any log files at all.
Temporary files used for logging in this case are
removed immediately after test's exit. Logs for
failing tests are still dumped to console.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -690,8 +690,10 @@ def main():
   parser = default_options_parser()
   (options, binaries) = parser.parse_args()
 
-  if not options.output_dir or not os.path.isdir(options.output_dir):
-    options.output_dir = None
+  if (options.output_dir is not None and
+      not os.path.isdir(options.output_dir)):
+    parser.error('--output_dir value must be an existing directory, '
+                 'current value is "%s"' % options.output_dir)
 
   # Append gtest-parallel-logs to log output, this is to avoid deleting user
   # data if an user passes a directory where files are already present. If a
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -184,8 +184,10 @@ class Task(object):
 
   @staticmethod
   def _logname(output_dir, test_binary, test_name, execution_number):
+    # Store logs to temporary files if there is no output_dir.
     if output_dir is None:
-      (log_handle, log_name) = tempfile.mkstemp(prefix='tmpGP_', suffix=".log")
+      (log_handle, log_name) = tempfile.mkstemp(prefix='gtest_parallel_',
+                                                suffix=".log")
       os.close(log_handle)
       return log_name
 
@@ -687,11 +689,15 @@ def main():
 
   parser = default_options_parser()
   (options, binaries) = parser.parse_args()
+
+  if not options.output_dir or not os.path.isdir(options.output_dir):
+    options.output_dir = None
+
   # Append gtest-parallel-logs to log output, this is to avoid deleting user
   # data if an user passes a directory where files are already present. If a
   # user specifies --output_dir=Docs/, we'll create Docs/gtest-parallel-logs
   # and clean that directory out on startup, instead of nuking Docs/.
-  if options.output_dir is not None:
+  if options.output_dir:
     options.output_dir = os.path.join(options.output_dir,
                                       'gtest-parallel-logs')
 
@@ -714,7 +720,7 @@ def main():
   assert len(unique_binaries) == len(binaries), (
       "All test binaries must have an unique basename.")
 
-  if options.output_dir is not None:
+  if options.output_dir:
     # Remove files from old test runs.
     if os.path.isdir(options.output_dir):
       shutil.rmtree(options.output_dir)
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -27,7 +27,6 @@ import tempfile
 import thread
 import threading
 import time
-import zlib
 
 if sys.platform == 'win32':
   import msvcrt
@@ -185,6 +184,11 @@ class Task(object):
 
   @staticmethod
   def _logname(output_dir, test_binary, test_name, execution_number):
+    if output_dir is None:
+      (log_handle, log_name) = tempfile.mkstemp(prefix='tmpGP_', suffix=".log")
+      os.close(log_handle)
+      return log_name
+
     log_name = '%s-%s-%d.log' % (Task._normalize(os.path.basename(test_binary)),
                                  Task._normalize(test_name), execution_number)
 
@@ -290,6 +294,9 @@ class FilterFormat(object):
     self.stdout_lock = threading.Lock()
 
   def move_to(self, destination_dir, tasks):
+    if self.output_dir is None:
+      return
+
     destination_dir = os.path.join(self.output_dir, destination_dir)
     os.makedirs(destination_dir)
     for task in tasks:
@@ -320,6 +327,9 @@ class FilterFormat(object):
           "[%d/%d] %s returned/aborted with exit code %d (%d ms)"
           % (self.finished_tasks, self.total_tasks, task.test_name,
              task.exit_code, task.runtime_ms))
+
+    if self.output_dir is None:
+      os.remove(task.log_file)
 
   def log_tasks(self, total_tasks):
     self.total_tasks += total_tasks
@@ -624,8 +634,7 @@ def default_options_parser():
   parser = optparse.OptionParser(
       usage = 'usage: %prog [options] binary [binary ...] -- [additional args]')
 
-  parser.add_option('-d', '--output_dir', type='string',
-                    default=tempfile.gettempdir(),
+  parser.add_option('-d', '--output_dir', type='string', default=None,
                     help='Output directory for test logs. Logs will be '
                          'available under gtest-parallel-logs/, so '
                          '--output_dir=/tmp will results in all logs being '
@@ -682,7 +691,9 @@ def main():
   # data if an user passes a directory where files are already present. If a
   # user specifies --output_dir=Docs/, we'll create Docs/gtest-parallel-logs
   # and clean that directory out on startup, instead of nuking Docs/.
-  options.output_dir = os.path.join(options.output_dir, 'gtest-parallel-logs')
+  if options.output_dir is not None:
+    options.output_dir = os.path.join(options.output_dir,
+                                      'gtest-parallel-logs')
 
   if binaries == []:
     parser.print_usage()
@@ -703,16 +714,17 @@ def main():
   assert len(unique_binaries) == len(binaries), (
       "All test binaries must have an unique basename.")
 
-  # Remove files from old test runs.
-  if os.path.isdir(options.output_dir):
-    shutil.rmtree(options.output_dir)
-  # Create directory for test log output.
-  try:
-    os.makedirs(options.output_dir)
-  except OSError as e:
-    # Ignore errors if this directory already exists.
-    if e.errno != errno.EEXIST or not os.path.isdir(options.output_dir):
-      raise e
+  if options.output_dir is not None:
+    # Remove files from old test runs.
+    if os.path.isdir(options.output_dir):
+      shutil.rmtree(options.output_dir)
+    # Create directory for test log output.
+    try:
+      os.makedirs(options.output_dir)
+    except OSError as e:
+      # Ignore errors if this directory already exists.
+      if e.errno != errno.EEXIST or not os.path.isdir(options.output_dir):
+        raise e
 
   timeout = None
   if options.timeout is not None:
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -315,8 +315,8 @@ class TestTask(unittest.TestCase):
       return 'C:\\' if sys.platform == 'win32' else '/'
 
     self.assertEqual(
-      'bin-Test_case-100.log',
-      gtest_parallel.Task._logname('', 'bin', 'Test.case', 100))
+      './bin-Test_case-100.log',
+      gtest_parallel.Task._logname('.', 'bin', 'Test.case', 100))
 
     self.assertEqual(
       os.path.join('..', 'a', 'b', 'bin-Test_case_2-1.log'),
@@ -342,6 +342,7 @@ class TestTask(unittest.TestCase):
                                    os.path.join(root(), 'c', 'd', 'bin'),
                                    'Test.case', 1))
 
+  def test_logs_to_temporary_files_without_output_dir(self):
     log_file = gtest_parallel.Task._logname(None, None, None, None)
     self.assertEqual(tempfile.gettempdir(), os.path.dirname(log_file))
     os.remove(log_file)
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import contextlib
-import gtest_parallel
 import os.path
 import random
 import shutil
@@ -21,6 +20,8 @@ import sys
 import tempfile
 import threading
 import unittest
+
+import gtest_parallel
 
 from gtest_parallel_mocks import LoggerMock
 from gtest_parallel_mocks import SubprocessMock
@@ -287,6 +288,14 @@ class TestTestTimes(unittest.TestCase):
 
           times.write_to_file(save_file)
 
+        self.assertEqual(
+          1000,
+          times.get_test_time('{}-{}'.format(path_to_binary, cnt),
+                              'TestFoo.testBar'))
+        self.assertIsNone(
+          times.get_test_time('{}-{}'.format(path_to_binary, cnt),
+                              'baz'))
+
       t = threading.Thread(target=test_times_worker)
       t.start()
       return t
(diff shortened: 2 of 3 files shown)


gtest_parallel.py
```diff
@@ -185,6 +184,11 @@ def _normalize(string):
 
   @staticmethod
   def _logname(output_dir, test_binary, test_name, execution_number):
+    if output_dir is None:
+      (log_handle, log_name) = tempfile.mkstemp(prefix='tmpGP_', suffix=".log")
```
pbos: Would it be reasonable to prefix with `tmp_gtest_parallel_`, or just `gtest_parallel_` or is that too long? `tmpGP` doesn't really read as `gtest-parallel` if you don't know where the file comes from to begin with. Omitting `tmpGP_` is also fine, but I like the prefix idea.

gtest_parallel.py
```diff
@@ -185,6 +184,11 @@ def _normalize(string):
 
   @staticmethod
   def _logname(output_dir, test_binary, test_name, execution_number):
+    if output_dir is None:
```
pbos: Comment here saying `# Store logs to temporary files while running if there's no output_dir.` or something like that.

gtest_parallel.py
```diff
@@ -703,16 +714,17 @@ def main():
...
-  if os.path.isdir(options.output_dir):
-    shutil.rmtree(options.output_dir)
-  # Create directory for test log output.
-  try:
-    os.makedirs(options.output_dir)
-  except OSError as e:
-    # Ignore errors if this directory already exists.
-    if e.errno != errno.EEXIST or not os.path.isdir(options.output_dir):
-      raise e
+  if options.output_dir is not None:
```
pbos: This can be `if options.output_dir:` right?

gtest_parallel_tests.py
```diff
@@ -333,6 +342,133 @@ def root():
                                    os.path.join(root(), 'c', 'd', 'bin'),
                                    'Test.case', 1))
 
+    log_file = gtest_parallel.Task._logname(None, None, None, None)
```
pbos: Can you make this a separate case? Just `def test_logs_to_temporary_files_without_output_dir:` ?

pbos: Can't comment on the right line (because GitHub), but line 319, isn't this wrong? If we log to empty string I think that should be the same as None, I think `--output_dir=` should probably be rejected or equivalent to omitting `--output_dir`. I think `--output_dir=.` should be used for the behavior on that line. (This is the `_logname('',` call.)

gtest_parallel.py
pbos: I think '' shouldn't be accepted as a command-line parameter.

gtest_parallel.py
volo-zyko: changed to `gtest_parallel_`

gtest_parallel.py
volo-zyko: Should we check that `output_dir` exists and reset it to `None` if it doesn't?

gtest_parallel_tests.py
volo-zyko: > line 319, isn't this wrong?

I don't see anything wrong with it. However, it can never happen on a normal run. Even if `--output_dir` is `''` on the command line it changes to `gtest-parallel-logs` in `main()`

gtest_parallel.py
pbos: I think we should reject the parameter it if the directory doesn't exist, can we validate this after parser.parse_args()?

gtest_parallel_tests.py
pbos: If it can't ever happen I think `'.'` is a better test case as we can definitely provide that argument.

pbos:
Thanks! Sorry for the slow response, I ended up not checking this out a lot over Christmas and then it kinda just kept getting punted.

gtest_parallel.py
```diff
@@ -678,11 +689,19 @@ def main():
 
   parser = default_options_parser()
   (options, binaries) = parser.parse_args()
+
+  if (options.output_dir is not None and
+      not os.path.isdir(options.output_dir)):
+    parser.error('--output_dir value must be an existing directory, '
```
oprypin, line 695: Here it says `output_dir` must exist

gtest_parallel.py
```diff
@@ -703,16 +722,17 @@ def main():
...
-    # Ignore errors if this directory already exists.
-    if e.errno != errno.EEXIST or not os.path.isdir(options.output_dir):
-      raise e
+  if options.output_dir:
+    # Remove files from old test runs.
+    if os.path.isdir(options.output_dir):
+      shutil.rmtree(options.output_dir)
+    # Create directory for test log output.
+    try:
+      os.makedirs(options.output_dir)
```
oprypin, line 731: Here it attempts to create `output_dir` but from a quick look this code can never be reached.

oprypin (commented):
This causes breakage for users that rely on it creating a directory to write logs into.

One such case is https://webrtc-review.googlesource.com/54722
Of course, there it is not causing immediate problems and is probably easily fixable, but I thought it's worth mentioning.

gtest_parallel.py
pbos: It's reachable if the original `output_dir` exists. At this point output_dir should be (original_output_dir)/gtest-parallel-logs. This should probably have a comment. Can you fix the missing base directory on your end?

gtest_parallel.py
```diff
@@ -678,11 +689,19 @@ def main():
...
+    parser.error('--output_dir value must be an existing directory, '
+                 'current value is "%s"' % options.output_dir)
+
   # Append gtest-parallel-logs to log output, this is to avoid deleting user
   # data if an user passes a directory where files are already present. If a
   # user specifies --output_dir=Docs/, we'll create Docs/gtest-parallel-logs
   # and clean that directory out on startup, instead of nuking Docs/.
-  options.output_dir = os.path.join(options.output_dir, 'gtest-parallel-logs')
+  if options.output_dir:
+    options.output_dir = os.path.join(options.output_dir,
```
oprypin, line 703: So this is what you're talking about.
Yeah, wouldn't normally expect `options` to change.

Feel free to disregard my other comments

pbos: Yeah sorry this is a little hard to read, we update the existing variable to prevent the rest of the code from accidentally reading the old value and mixing up `output_dir` with `output_dir/gtest-parallel-logs`.

pbos:
@volo-zyko after running this without `--output_dir` I get the following on Windows:

```
C:\src\gtest-parallel (master)>python gtest-parallel \src\chromium\src\out\Default\unit_tests.exe
[0/8623] Running tests...
[1/8623] DownloadSuggestionsProviderTest.ShouldFetchOfflinePageDownloadsOnStartup (2536 ms)
[2/8623] DownloadSuggestionsProviderTest.ShouldFetchAssetDownloadsOnStartup (2555 ms)
Exception in thread Thread-7:
Traceback (most recent call last):
  File "C:\src\depot_tools\win_tools-2_7_6_bin\python\bin\lib\threading.py", line 810, in __bootstrap_inner
    self.run()
  File "C:\src\depot_tools\win_tools-2_7_6_bin\python\bin\lib\threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 610, in __call__
    task_manager.run_task(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 267, in run_task
    self.__register_exit(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 249, in __register_exit
    self.logger.log_exit(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 334, in log_exit
    os.remove(task.log_file)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\pbos\\appdata\\local\\temp\\gtest_parallel_9uyg2j.log'

[3/8623] DownloadSuggestionsProviderTest.ShouldReplaceRemovedItemWithNewData (2573 ms)
Exception in thread Thread-4:
Traceback (most recent call last):
  File "C:\src\depot_tools\win_tools-2_7_6_bin\python\bin\lib\threading.py", line 810, in __bootstrap_inner
    self.run()
  File "C:\src\depot_tools\win_tools-2_7_6_bin\python\bin\lib\threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 610, in __call__
    task_manager.run_task(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 267, in run_task
    self.__register_exit(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 249, in __register_exit
    self.logger.log_exit(task)
  File "C:\src\gtest-parallel\gtest_parallel.py", line 334, in log_exit
    os.remove(task.log_file)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\pbos\\appdata\\local\\temp\\gtest_parallel_2mrprc.log'
```

I don't see how this is possible, and it only happens for the N first test runs or so. It kind of makes me think that the `os.close` call is insufficient for the python process to have dropped all handles to the `mkstemp` file. Should we consider rewriting it it with output_dir as `mkdtemp` and then appending gtest-parallel-logs etc. as usual, and remove the directory tree in the end? Anything else you can think of? I think this is trying to do the right thing.

pbos:
At this point it feels like `subprocess.Popen` is still holding on to a handle or something. Shot in the dark. One way that worked as a workaround was to put all os.remove() objects in a list and remove those at the end of execution, but that feels super hacky.

volo-zyko:
I'll take a look at this. I saw such errors in interrupted tests and wanted to suppress them as it looks like the files are deleted at the end. But first, let's see why it happens.

pbos:
@volo-zyko Thanks! I tried hunting it down as well. Invoking SysInternals' `handle.exe` before removing the file didn't find the offending process fast enough, so I'm not sure if this is Popen's `CreateProcess` call still holding a handle to the file. Which platform(s) are you seeing this on? A quick workaround if we can't get to the bottom of this is to use `mkdtemp` and generate short lognames like `"log_%d.txt" % test_invocation_number` or something (to avoid the long-logname problem), and `shutil.rmtree` at the end of invocation. This would at least contain the logs in one folder, if anything crashes or `kill -9` or similar.

Thinking about that, maybe using numbered logs are preferable with `--output_dir` specified (since testname + dir can easily hit your `140` limit).  Maybe using `log_%d.txt % task_id` is preferable even with `--output_dir` specified? We can then move it to `{passed,failed,interrputed}passed/log_%d.txt % task_id` and it should be short enough in all cases. We just need to make sure that the `%d` here after moving the log files match the `[%d / num_runs]` reported in the console so that a user can find the right logs. Maybe the `%d` reported is not the same as `task_id` in which case that should be used in place of `task_id`.

@oprypin I don't think our discussion here affects WebRTC buildbots since they specify `--output_dir`, let me know if that's not the case and we can land a short-term solution for this. I still think the bots should create `--output_dir` though.

volo-zyko:
> Which platform(s) are you seeing this on?

Surely it was Windows. Windows 7 to be specific.

Concerning proposed workarounds. Both of them will complicate things, so before resorting to any of them I want to investigate if there is anything simpler.

volo-zyko:
Well, today I took a look at this problem and from what I can see it originates from the internals of `subprocess.Popen`. However, I'm not completely sure who is to blame - Python or Windows.

Here is how it runs. We obtain the first file handle to a log file when `gtest_parallel` opens it in `Task.run`. Then `subprocess.Popen` duplicates with Windows API `DuplicateHandle` this handle to stdout and stderr for a child process, hence now we have 3 open file handles. Then it starts the child process and as far as I understand the child (or rather Windows) also creates 2 file handles to that file. After the child is successfully spawned parent process (inside `subprocess.Popen`) closes own duplicates for child stdout and stderr. Then inside `Task.run` we wait for the child, which essentially does `WaitForSingleObject`. Once it's done, which means that the child process has terminated and presumably closed own stdout and stderr handles pointing to the log file, `gtest_parallel` closes own log file handle. That must be it but not quite. There is a race which is clearly visible in ProcessMonitor between `WaitForSingleObject` and actually running child processes. In some `gtest_parallel` runs `WaitForSingleObject` returns too early and we attempt to remove log file which is still used by some child process.

Just a note. Inserting `time.sleep(0.01)` at the end of `Task.run` did help but it doesn't look like a good solution.

See [here](https://github.com/python/cpython/blob/5dc275a2c00839cce4cd717f64fc9456e05d5af3/Lib/subprocess.py#L334) details on `subprocess.Popen` implementation.

pbos:
Gross, not sure about what to do about that. To mitigate this I propose that we either:

1. Replace `mkstemp` with `mkdtemp`.
1. Generate shorter log names even when `--output_dir` is run, with names matching the order in which they finished (since that matches the `[476/900] sometest.sometest` print.

or:

1. Store the logname in `Task`.
1. Loop through all tasks at the end of execution and remove them (instead of closing them as the processes finish).

What do you think about either of those approaches? The latter is simpler but the former lets you have `--output_dir=foo` on systems that have filepath length restrictions, which I think would be a benefit.

volo-zyko:
Well, both solution don't reliably fix the afore-mentioned race condition. If we start cleanup immediately after termination of the last test then there is a chance that it's still running. As for me if such a low level thing as `WaitForSingleObject` can prematurely signal process termination while it's still running then the best thing that we can do is to make sure that the log file is really removed, i.e. istead of plain `os.remove(task.log_file)` in `FilterFormat.log_exit` we should do something like this:

```python
while True:
  if not os.path.isfile(task.log_file):
    break

  try:
    os.remove(task.log_file)
  except OSError:
    pass
```

oprypin:
Just got 

> WindowsError: [Error 32] The process cannot access the file because it is being used by another process 

https://build.chromium.org/p/client.webrtc/builders/Win%20%28more%20configs%29/builds/1449/steps/peerconnection_unittests
And earlier
https://build.chromium.org/p/client.webrtc/builders/Win%20%28more%20configs%29/builds/1390/steps/peerconnection_unittests

pbos:
Thanks @volo-zyko, that's not very nice :( but I think it might be the best we got. I think we'll need to add a timeout too. I'll look into it.

Did you remove `--output_dir`? The file referenced here is generated using `mkstemp` which shouldn't happen with `--output_dir` present. I guess we have to add something like the hacky workaround. I'll take a look at it today, but kinda gross.

I think I'll only add the workaround for Windows specifically, please let me know if you see anything on other platforms.

pbos:
The workaround is now on all platforms because I wanted to reference `errno.ENOENT` in the rare case the file gets removed from under our feet.

pbos:
@volo-zyko I wonder if this will repro with N workers running:

`msktemp`, `task = subprocess.popen`, `task.wait`, `os.remove`

Without any sleep in between. WDYT? I'd be interesting in reporting that upstream even for 2.7.

volo-zyko:
Well, the following snippet reproduces the problem:

```python
import os
import subprocess
import time
import thread
import threading


def spawn_workers(workers_count):
    def worker_fn():
        file_name = str(thread.get_ident()) + '.log'
        print 'Logging to %s\n' % file_name

        # command = ['timeout', '3']
        command = ['ping', '-n', '3', '127.0.0.1']
        with open(file_name, 'w') as file_obj:
            task = subprocess.Popen(command, stdout=file_obj, stderr=file_obj)
            task.wait()

        # time.sleep(0.1)
        print 'Removing %s\n' % file_name
        os.remove(file_name)

    def start_daemon(func):
        t = threading.Thread(target=func)
        t.daemon = True
        t.start()
        return t

    workers = []
    try:
        workers = [start_daemon(worker_fn) for _ in range(workers_count)]
    finally:
        for worker in workers:
            worker.join()


if __name__ == '__main__':
    spawn_workers(4)
```

Here is the output:

```
Logging to 7300.log
Logging to 6040.log

Logging to 6832.log


Logging to 308.log

Removing 7300.log

Removing 6832.log

Removing 308.log
Removing 6040.log
E
xception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "C:\Python27\lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "race.py", line 21, in __call__
    os.remove(file_name)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: '7300.log'
```

It's interesting to note that if the number of workers is small (1 or 2) then the problem does not reproduce even without `sleep`. With more workers the problem can be fixed by adding short sleep. However, if the commmand to run is `timeout` then even `sleep(5)` doesn't help. `timeout` command is strange and probably better use `ping` for testing.

pbos:
Sorry for the delay, I filed an upstream bug here: https://bugs.python.org/issue33369

PR #61, opened 2018-03-28
Add Python3 compatibility

maflcko:
This adds python3 compatibility such that the tool runs with both python2.7 and python3

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -11,8 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import cPickle
 import errno
+from functools import total_ordering
 import gzip
 import json
 import multiprocessing
@@ -24,9 +24,18 @@ import signal
 import subprocess
 import sys
 import tempfile
-import thread
 import threading
 import time
+
+if sys.version_info.major >= 3:
+    long = int
+    import _pickle as cPickle
+    import _thread as thread
+else:
+    import cPickle
+    import thread
+
+from pickle import HIGHEST_PROTOCOL as PICKLE_HIGHEST_PROTOCOL
 
 if sys.platform == 'win32':
   import msvcrt
@@ -140,6 +149,7 @@ def get_save_file_path():
     return os.path.join(os.path.expanduser('~'), '.gtest-parallel-times')
 
 
+@total_ordering
 class Task(object):
   """Stores information about a task (single execution of a test).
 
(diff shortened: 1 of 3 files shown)


gtest_parallel.py
```diff
@@ -11,23 +11,34 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import cPickle
+try:
+    import _pickle as cPickle
```
pbos: Seems to me like `cPickle` is the reasonable default, can we import that one?

gtest_parallel.py
```diff
@@ -487,7 +505,7 @@ def __init__(self, save_file):
     # Discard saved times if the format isn't right.
     if type(times) is not dict:
       return
-    for ((test_binary, test_name), runtime) in times.items():
+    for ((test_binary, test_name), runtime) in list(times.items()):
```
pbos: Why do these need to be converted to lists?

gtest_parallel_mocks.py
```diff
@@ -157,7 +157,7 @@ def __call__(self, command, **kwargs):
     self.last_invocation = command
     binary = command[0]
     test_list = []
-    tests_for_binary = sorted(self._test_data.get(binary, {}).iteritems())
+    tests_for_binary = sorted(self._test_data.get(binary, {}).items())
```
pbos, line 160: I don't understand the list conversions everywhere, are iterators through iteritems broken or unsupported in Python3?

gtest_parallel.py
pbos: Same here and below.

gtest_parallel.py
```diff
@@ -11,23 +11,34 @@
...
+from pickle import HIGHEST_PROTOCOL as PICKLE_HIGHEST_PROTOCOL
 import re
 import shutil
 import signal
 import subprocess
 import sys
 import tempfile
-import thread
+try:
+    import _thread as thread
```
pbos: Actually, can you put conditional imports under `sys.version_info.major` (and corresponding `else:`) instead of try blocks?

pbos (commented):
Sorry that I got to this slow, I don't see why we need to do conversions to lists everywhere. Is that busted in Python3 for some reason?

gtest_parallel_mocks.py
pbos: ok, I read up a bit on this, as iteritems are no longer supported I think this instance is fine but are the other list conversions necessary?

maflcko: Good catch. The others are not required, since they are only used in for loops, which can handle the iterators.

gtest_parallel_tests.py
```diff
@@ -109,7 +109,7 @@ def execute_tasks(self, tasks, retries, expected_exit_code):
     for test_id, expected in tasks:
       task = task_mock_factory.get_task(test_id)
       task_manager.run_task(task)
-      expected['execution_number'] = range(len(expected['exit_code']))
+      expected['execution_number'] = list(range(len(expected['exit_code'])))
```
pbos, line 112: last nag I think, is this list conversion necessary?

maflcko: Yes, the tests use `assertListEqual` and would otherwise fail.

PR #62, opened 2018-04-30
XML output

v-arbatov:
I often use google test library together with Jenkins and I like running them in parallel. I find gtest-parallel script very useful but somehow it doesn't have XML output compatible with Jenkins JUnit plugin. To get things working out of the box I refactored script to allow different test results collectors and added XML results collector.
Similar to `--dump_json_test_results=DUMP_JSON_TEST_RESULTS` there is `--dump_xml_test_results=DUMP_XML_TEST_RESULTS` added.
Additional option `--dump_xml_all_logs` controls if output logs of all tests are saved in XML, I often need this included and this is even better than gtest XML output because gtest-parallel captures whole and all output (in my case gtest may run under simulator which may report additional info).

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,3 +97,15 @@ to avoid mixing old stale logs with new ones. This will discard any logs from
 ongoing invocations, or otherwise fail (throw) because the files are in use
 (depending on platform). This is not enforced by a lock file as cleaning these
 up would require unnecessary user intervention.
+
+## XML Output
+
+This script is able to save test results in XML format. Use
+`--dump_xml_test_results=` to enable XML output and specify file name.
+By default standard error and output are saved in XML for failed tests.
+Include `--dump_xml_all_logs` in command line to save output of all logs.
+
+For example:
+
+    $ ./gtest-parallel path/to/binary... --dump_xml_test_results=example.xml --dump_xml_all_logs
+
--- a/README.md
+++ b/README.md
@@ -91,12 +91,16 @@ this permits such binaries to partially benefit from parallel execution.
 
 ## XML Output
 
-This script is able to save test results in XML format. Use
-`--dump_xml_test_results=` to enable XML output and specify file name.
-By default standard error and output are saved in XML for failed tests.
-Include `--dump_xml_all_logs` in command line to save output of all logs.
+This script supports saving test results as Google Test XML. Use
+`--gtest_output=FORMAT[:DIRECTORY_PATH/|:FILE_PATH]` parameter to enable
+this feature. FORMAT is either `xml` or `xml-full`. FILE_PATH defaults to
+test_detail.xml.
+
+Standard error and output are saved in `<system-out>` tag under `<testcase> tag
+in addition to other Google Test XML data. In `xml` format only failed tests
+have output saved. In `xml-full` format output is saved for all tests.
 
 For example:
 
-    $ ./gtest-parallel path/to/binary... --dump_xml_test_results=example.xml --dump_xml_all_logs
+    $ ./gtest-parallel path/to/binary... --gtest_output=xml:example.xml
 
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -15,6 +15,7 @@ import cPickle
 import errno
 import gzip
 import json
+import xml.etree.cElementTree as ET
 import multiprocessing
 import optparse
 import os
@@ -28,6 +29,7 @@ import thread
 import threading
 import time
 import zlib
+import datetime
 
 if sys.platform == 'win32':
   import msvcrt
@@ -198,6 +200,7 @@ class Task(object):
         self.exit_code = sigint_handler.wait(task)
       except sigint_handler.ProcessWasInterrupted:
         thread.exit()
+
     self.runtime_ms = int(1000 * (time.time() - begin))
     self.last_execution_time = None if self.exit_code else self.runtime_ms
 
@@ -243,9 +246,8 @@ class TaskManager(object):
     self.logger.log_exit(task)
     self.times.record_test_time(task.test_binary, task.test_name,
                                 task.last_execution_time)
-    if self.test_results:
-      self.test_results.log(task.test_name, task.runtime_ms,
-                            "PASS" if task.exit_code == 0 else "FAIL")
+    for sink in self.test_results:
+      sink.log(task)
 
     with self.lock:
       self.started.pop(task.task_id)
@@ -356,7 +358,7 @@ class FilterFormat(object):
     self.out.flush_transient_output()
 
 
-class CollectTestResults(object):
+class CollectJSONTestResults(object):
   def __init__(self, json_dump_filepath):
     self.test_results_lock = threading.Lock()
     self.json_dump_file = open(json_dump_filepath, 'w')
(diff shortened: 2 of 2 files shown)


v-arbatov:
I signed it!

gtest_parallel.py
```diff
@@ -402,27 +403,108 @@ def __init__(self, json_dump_filepath):
...
+      group = self.test_results['children'].setdefault(group_name, {
+        "tag": "testsuite",
+        "attrib": {
+          "name": group_name,
+          "disabled": 0,
+          "errors": 0,
+        },
+        "children": {},
+      })
+      assert test_name not in group['children'], "Multiple runs are not supported with XML test results output"
```
pbos: This needs to be verified earlier as part of command line validation (keep the assert though). See `parser.error` usage in `main()` for instance.

README.md
```diff
@@ -88,3 +88,15 @@ For such binaries where test cases are independent, `gtest-parallel` provides
 `--serialize_test_cases` that runs tests within the same test case sequentially.
 While generally not providing as much speedup as fully parallel test execution,
 this permits such binaries to partially benefit from parallel execution.
+
+## XML Output
+
+This script is able to save test results in XML format. Use
+`--dump_xml_test_results=` to enable XML output and specify file name.
+By default standard error and output are saved in XML for failed tests.
+Include `--dump_xml_all_logs` in command line to save output of all logs.
```
pbos: Is there a downside to always dumping all log output? I'd prefer not adding more options here (and avoid a new flag specifically per format).

README.md
```diff
@@ -88,3 +88,15 @@ For such binaries where test cases are independent, `gtest-parallel` provides
 `--serialize_test_cases` that runs tests within the same test case sequentially.
 While generally not providing as much speedup as fully parallel test execution,
 this permits such binaries to partially benefit from parallel execution.
+
+## XML Output
+
+This script is able to save test results in XML format. Use
```
pbos: What XML format? Where does it come from? What's the specification? If this is a specific JUnit format, then please say so and link to the spec. Additionally I think we should call it something like `--dump_junit_test_results` instead of calling it generic "XML".

Also, what do you think about having something like `--output_format=junit` and `--output=results.xml` ? That way we could add new formats without adding new flags. Do you see a need for having multiple concurrent sinks?

gtest_parallel.py
```diff
@@ -754,9 +840,12 @@ def main():
   if options.timeout is not None:
     timeout = threading.Timer(options.timeout, sigint_handler.interrupt)
 
-  test_results = None
+  test_results = []
```
pbos: I think we can have one sink instead of appending. Especially if you're OK with the suggested output format flag.

gtest_parallel.py
```diff
@@ -684,6 +766,10 @@ def default_options_parser():
                     help='Saves the results of the tests as a JSON machine-'
                          'readable file. The format of the file is specified at '
                          'https://www.chromium.org/developers/the-json-test-results-format')
+  parser.add_option('--dump_xml_test_results', type='string', default=None,
+                    help='Saves the results of the tests as a XML')
```
pbos: Same here, this needs to document what the format actually is.

pbos:
Thanks. I think this is a good idea in general, but I think we should document the format and consider not needing a new flag for every added format.

Also, could you please add some tests that verify that test results are formatted and merged correctly?

pbos (changes requested):

gtest_parallel.py
v-arbatov: Yes, failing earlier is better. I'm not sure what turns on multiple runs, I'll take a closer look.

README.md
v-arbatov: I think the downside is potentially large XML file and the need to parse it later. Personally, I save all logs just in case I want to take a look at history but I can imagine people disagree that this is needed.

README.md
v-arbatov: Well, the format resembles the one described here: https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md
Intention to be parseable by JUnit plugin in Jenkins. I can refer to that ocument with outlining what was added.
If same switch is used for multiple formats, can it follow same approach as gtest and specify format right in the command, for example `--dump_test_results=junit:filename` with `--dump_junit_test_results` preserved for backward compatibility? I think I could have, say, `xml` and `xml-full` formats instead of separate format specific flag `--dump_xml_all_logs`.
I also think that only one test results sink is needed. In my code concurrent sink appeared from independent --dump_ flags.

pbos: Looks like there already is precedence for an existing flag, per that page: "To generate the XML report, set the GTEST_OUTPUT environment variable or the --gtest_output flag to the string "xml:_path_to_output_file_", which will create the file at the given location."

If you're reusing a gtest flag with the same semantics as that gtest flag I think that lessens the burden of us to document the format / behavior (because it does what it's expected to do). We should still link to gtest documentation appropriately here.

So implementing `--gtest_output=xml:path_to_file.xml` makes sense to me here. There's no need to implement any other flags, skip `dump_xml_all_logs` until there's actually a need for it (and we figure out how we want it implemented). I think you can filter an xml file to drop passed results fairly cheaply and the cost of creating and storing it needs to be substantial for us to consider the all-logs argument (it should do the same as `--gtest_output` on the gtest binary does by default).

I'm not sure the existing JSON format is the same as the gtest json format. Unless you know this format is the same that gtest uses are the same, do not implement `--gtest_output=json:` using it. If they are the same then having it use `--gtest_output` is better as the users don't need to know the flag (but the old one needs to be kept a bit for compatibility).

v-arbatov: I didn't want to use `--gtest_output` flag because generated XML is not exactly the same as the one produced by gtest. The difference is mainly in failures messages and captured standard output.
Let's try the following: use `--gtest_output` for xml output, generate xml file for each individual test, parse them and aggregate results into single xml output as user requested. I agree, it would be easier to document then. It will take some time for me to implement though as I'm doing this on my spare time.
I still need to add standard output to generated XML tests, I'm not happy with what I'm getting from gtest by default. Perhaps under `xml-full` format.
I'm not going to implement JSON format support. In fact, I didn't know it's supported now in gtest.

pbos: Are the formats compatible (are you using the same tags, headers etc.) so that a parser for --gtest_output=xml:foo would read something reasonable from our output?

If so I think you can add --gtest_output and have it generate what you want instead of merging --gtest_output from the binaries. Is there a reasonable heading in the existing format where you can put stdout?

No rush either way btw, I'm running this in my spare time as well and know how things can slip.

pbos: How do the failure messages and and stdout reported in the formats differ?

v-arbatov: In gtest failure message contains same text as printed in gtest. In my XML it's just a message that test exited with non zero status. I don't like this, I'd rather have original gtest failure mesage. Stdout is not reported in gtest but if added as <system-out> subtag of testcase then Jenkins JUnit parser will understand it.
I noticed also that user can put custom tags into XML output in gtest right from his tests, gtest provides API for this.
As I mentioned above, I think a better solution is to capture XML output of individual tests and merge them into single report. If test crashed and didn't produce XML report then I can create dummy one. This is not much harder to implement. I'm trying this approach now.

pbos: Sounds good, I think the less `gtest-parallel` has to invent in general the better (using the existing XMLs sound good to me). Keep `--repeat` in mind, the existing gtest xml output might provide guidance as it likely handles `--gtest_repeat`.

v-arbatov:
I reworked XML output. Now --gtest_output parameter is used, for XML only. I looked how repeated test runs are handled in gtest xml output and apparently, only last result is saved, even if test failed on previous run. I decided to keep/save failed tests during merging.
Regarding tests, I wonder, how do you see this? Ideally it should run some gtest executable. Perhaps, checkout googltest, build examples there and run?

README.md
```diff
@@ -88,3 +88,19 @@ For such binaries where test cases are independent, `gtest-parallel` provides
...
+## XML Output
+
+This script supports saving test results as Google Test XML. Use
+`--gtest_output=FORMAT[:DIRECTORY_PATH/|:FILE_PATH]` parameter to enable
+this feature. FORMAT is either `xml` or `xml-full`. FILE_PATH defaults to
+test_detail.xml.
+
+Standard error and output are saved in `<system-out>` tag under `<testcase> tag
+in addition to other Google Test XML data. In `xml` format only failed tests
+have output saved. In `xml-full` format output is saved for all tests.
```
pbos, line 101: This seems to diverge from `--gtest_output=xml` without `gtest-parallel`, they should match behavior whenever possible: `--gtest_output=xml` should have the same behavior as running `--gtest_output=xml` on the original gtest binary and include failures.

Separately, are both of these modes or a "xml-light" mode really required? Can't you filter out passing tests from the full logs as a separate script that's not part of `gtest-parallel`? I don't want to add `--gtest_output` modes that upstream gtest doesn't support.

pbos:
Sorry for the slow response cycle, this sat as a pending comment for months without me noticing. :(

spiricn:
Hello @v-arbatov @pbos,

Is this feature/PR officially dead or has there been any activity elsewhere that I'm not aware of ?

pbos:
Dead until resurrected. You might have some luck applying their patches.

Thinking out loud I'd wonder if `dump_json_test_results` would be translatable to something useful or if gtest (non-parallel) binaries' xml output could be merged reasonably without adding too much code (I'm concerned about growing the size of this as it's not really maintained), but I haven't touched this in a long while.

v-arbatov:
Yeah, no further development going on.

I still run my version on the project I have been working at the time.
Given that gtest now supports parallelization natievly there is not much
incentive to build something outside of gtest itself. My case is special
because our gtest runs single threaded under simulator and scripting
outside of gtest allows to parallelize runs and speedup turnaround time.

spiricn:
> Given that gtest now supports parallelization natievly there is not much incentive to build something outside of gtest itself.

@v-arbatov I might have missed something, what do you mean gtest now supports parallelization natively ?

v-arbatov:
I apologise, it's been a while... I had impression that this happened but I
don't see this feature in googletest.

PR #63, opened 2018-07-07
fix py3 compatibility

tchaikov:
because subprocess.check_output() return bytes not str in python3, we
need to decode it before split('\n').

otherwise we will have failures like

93: Traceback (most recent call last):
93:   File "/home/kefu/dev/ceph/build/src/test/gtest-parallel/gtest-parallel", line 18, in <module>
93:     sys.exit(gtest_parallel.main())
93:   File "/home/kefu/dev/ceph/build/src/test/gtest-parallel/gtest_parallel.py", line 785, in main
93:     tasks = find_tests(binaries, additional_args, options, times)
93:   File "/home/kefu/dev/ceph/build/src/test/gtest-parallel/gtest_parallel.py", line 579, in find_tests
93:     for line in test_list.split('\n'):
93: TypeError: a bytes-like object is required, not 'str'

Signed-off-by: Kefu Chai <tchaikov@gmail.com>

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -14,6 +14,7 @@
 import errno
 from functools import total_ordering
 import gzip
+import io
 import json
 import multiprocessing
 import optparse
@@ -302,7 +303,14 @@ class FilterFormat(object):
   def __init__(self, output_dir):
     if sys.stdout.isatty():
       # stdout needs to be unbuffered since the output is interactive.
-      sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
+      if isinstance(sys.stdout, io.TextIOWrapper):
+        # workaround for https://bugs.python.org/issue17404
+        sys.stdout = io.TextIOWrapper(sys.stdout.detach(),
+                                      line_buffering=True,
+                                      write_through=True,
+                                      newline='\n')
+      else:
+        sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
 
     self.output_dir = output_dir
 
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -573,10 +573,16 @@ def find_tests(binaries, additional_args, options, times):
     except subprocess.CalledProcessError as e:
       sys.exit("%s: %s" % (test_binary, str(e)))
 
+    try:
+        test_list = test_list.split('\n')
+    except TypeError:
+        # subprocess.check_output() returns bytes in python3
+        test_list = test_list.decode(sys.stdout.encoding).split('\n')
+
     command += additional_args + ['--gtest_color=' + options.gtest_color]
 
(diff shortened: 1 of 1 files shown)


tchaikov:
I signed it!

gtest_parallel.py
```diff
@@ -576,7 +584,7 @@ def find_tests(binaries, additional_args, options, times):
     command += additional_args + ['--gtest_color=' + options.gtest_color]
 
     test_group = ''
-    for line in test_list.split('\n'):
+    for line in test_list.decode().split('\n'):
```
pbos: Would something like this instead work instead of using the IOWrapper?

```
if sys.version_info.major >= 3
  test_list = str(test_list, encoding='utf-8')
```

pbos: or `encoding=sys.stdout.encoding`, I'm not very familiar with this but it would be really nice if we didn't have to add the workaround above.

tchaikov:
@pbos ping?

pbos:
Sorry, apparently I had review comments that I forgot to publish, my bad. :(

gtest_parallel.py
tchaikov: i'd go with `encoding=sys.stdout.encoding`. the check for python version is distracting from developer's point of view, IMHO.

and i am not sure how `IOWrapper` comes into play here. are you talking about the `FilterFormat` change?

tchaikov:
@pbos fixed and repushed.  as an alternative approach, i am catching `TypeError` and do the decode in the catch block, hopefully it's a better fix.

gtest_parallel.py
```diff
@@ -302,7 +303,14 @@ class FilterFormat(object):
   def __init__(self, output_dir):
     if sys.stdout.isatty():
       # stdout needs to be unbuffered since the output is interactive.
-      sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
+      if isinstance(sys.stdout, io.TextIOWrapper):
```
pbos, line 306: I'd like to get rid of this block in virtually any way possible. Is there no way to get the bytes off of here without it?

tchaikov: @pbos i am afraid that this fix is the best way i have in my mind to have unbuffered stdout in python3 =(

tchaikov: @pbos is there anything i can do to progress this PR?

pbos: Sorry, I was swamped since last time and wanted to have a chance to repro the errors you got when running. I'm able to run `python3 gtest-parallel /path/to/foo` with your changes and were not able to before. It seems like previous PRs to fix Py3 compatibility either had vastly different environments from ours or never really tested their fixes (I had no idea this was completely broken).

Long story short, thanks! Merging.

PR #69, opened 2019-05-11
Adding setup.py to project.

kshithijiyer:
Signed-off-by: kshithijiyer <kshithij.ki@gmail.com>


pbos:
Hi, there are existing deployments so we can't make `gtest-parallel/` a subdirectory without breaking things. I also prefer keeping things flatter in projects of this size.

Adding a `version.py` / `setup.py` would sort of imply that those are kept to date. Can `pip` function without a version number supplied & have `version.py` removed? I'd like to not have a version numbering scheme for this project and would definitely not be trusted to keep it up to date.

Otherwise, can pip files reasonably live outside this repository? A separate installer repository would be able to check this one out as a dependency and use the `gtest-parallel/` subdirectory you wanted.

kshithijiyer:
A separate repository won't make sense. As far as `version.py` is concerned I can easily remove the code. The project structure can be kept flat as well, no issues on that one. Lastly, I just didn't understand why to remove `setup.py`? Could you please explain to me?

pbos:
Does a `setup.py` work properly (e.g. does `pip` work) without version numbers?

kshithijiyer:
> Does a `setup.py` work properly (e.g. does `pip` work) without version numbers?
Yes. It seems to be working fine.

pbos:
Sorry for the slow response. I'll be out for a little over a week starting tomorrow.

Looking at https://docs.python.org/2/distutils/setupscript.html#additional-meta-data marks `version` as required.

My deal here is that I don't want to commit to keeping version numbers up to date. Having `pip` packages sort of implies that they are also officially maintained.

Here's the scenario I want to avoid:

1. We commit a `setup.py` with say `version=0.1`.
1. Users can now install `gtest-parallel`, all is fine.
1. We commit new changes to `gtest-parallel`, but forget to update `version`.
1. A new user installs `gtest-parallel` with pip, and receives a different version of `gtest-parallel` @ version 0.1.
1. Existing users call `pip install --upgrade`, but the version number is the same as before.

I see three problems problems arise in this scenario:
Users don't agree on what version 0.1 is.
Users can't automatically update to the latest version because we haven't made a release. I don't want to commit to making releases.
Users think that the `pip` package will be kept up to date, while the only up-to-date version is `HEAD`.

kshithijiyer:
Should we close this pull request?

pbos:
Will do, thanks.

PR #71, opened 2019-08-18
Let an OS decide which Python version to use to run tests

kuzkry:
Scripts are python 3 ready, therefore it's better not to hard code python 2 which may not even be available on some systems.

### DIFF
--- a/gtest_parallel_tests.py
+++ b/gtest_parallel_tests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Copyright 2017 Google Inc. All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


kuzkry:
Btw. I also thought about adding a missing hashbang to *gtest_parallel.py*, which might be run without *gtest-parallel*. What do you think?

pbos:
Good cleanup, this matches `gtest-parallel` which imo is the "supported way" of running `gtest_parallel.py`, so I don't think we should add a shebang into `gtest_parallel.py`. All documentation mention `gtest-parallel` as the target to run.

kuzkry:
Yeah, then it would be superfluous. OK, cool.

PR #73, opened 2020-03-27
Use `additional_args` to list tests

vitaliy505:
Hello!

I found a small issue with my tests. It uses command line parameters to parametrize tests written with gtest. And it doesn't work correctly with gtest-parallel because these arguments aren't used when list tests. 
In general, default behaviour shouldn't be affected.


vitaliy505:
@googlebot I signed it!

PR #76, opened 2020-06-23
Limit log file name length to OS allowed limit

asomsiko:


gtest_parallel.py
```diff
@@ -44,6 +44,11 @@
   import fcntl
 
 
+# The log file names is truncated to 256 to overcome OS limitations
```
pbos: Can you move these constants into where they are used?

gtest_parallel.py
```diff
@@ -204,14 +209,18 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
-      return log_name
-
-    log_name = '%s-%s-%d.log' % (Task._normalize(os.path.basename(test_binary)),
-                                 Task._normalize(test_name), execution_number)
+                                                suffix='.log')
+    else:
+      suffix = '-%d.log' % execution_number
+      prefix = '%s-%s' % (Task._normalize(os.path.basename(test_binary)),
+                          Task._normalize(test_name))
+      namemax = os.statvfs('.').f_namemax if hasattr(os, 'statvfs') else NAMEMAX_DEFAULT
```
pbos: Can you have this be: max_prefix_length = be (f_filemax - space_for_random) else default?

gtest_parallel.py
```diff
@@ -204,14 +209,18 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
+      (log_handle, log_name) = tempfile.mkstemp(prefix=prefix[:namemax - MKSTEMP_RANDOM_LEN - len(suffix)],
```
pbos: Can you rewrite this so that it always tries to create the full file name and then falls back to create a truncated version? That way you can have full-length names even if statvfs isn't supported.

Something like:
* Try creating the file no matter what
* Retry but shortening it using statvfs (or defaults)

gtest_parallel.py
```diff
@@ -204,14 +209,18 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
+                                                dir=output_dir,
+                                                suffix=suffix)
+    os.close(log_handle)
```
pbos: This is a different change from before. Did we have a handle leak here?

gtest_parallel.py
```diff
@@ -44,6 +44,11 @@
   import fcntl
 
 
+# The log file names is truncated to 256 to overcome OS limitations
+MKSTEMP_RANDOM_LEN = 12  # length of mkstemp random string being added
+NAMEMAX_DEFAULT = 255  # Limit for log file name length
```
pbos: Can this be renamed DEFAULT_PREFIX_LENGTH and maybe = 240?

gtest_parallel.py
asomsiko: There was no handle leak. The `mkstemp` in this change is also used in the second case so `os.close` closes the handle for both

gtest_parallel.py
asomsiko: Sure, renamed and set to 240

gtest_parallel.py
asomsiko: Good idea. I've noticed one test insists that the file should be removed after the creation attempt here.

gtest_parallel.py
asomsiko: Cleaned this up

gtest_parallel.py
asomsiko: Sure. Done

gtest_parallel.py
```diff
@@ -204,14 +204,39 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
+    try:
+      # check if log can be created or exists already
+      if not os.path.exists(log_name):
+        os.makedirs(output_dir, exist_ok=True)
+        with open(log_name, 'w+'):
+          pass
+        os.remove(log_name)
+    except OSError as os_error:
+      # truncate file name if error is Errno 36: File name too long
+      if os_error.errno == 36:
```
pbos, line 229: Can this be `errno.ENAMETOOLONG` instead of the comment?

pbos: https://docs.python.org/3/library/errno.html#errno.ENAMETOOLONG

gtest_parallel.py
```diff
@@ -204,14 +204,39 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
-                                                suffix=".log")
+                                                suffix='.log')
       os.close(log_handle)
       return log_name
 
-    log_name = '%s-%s-%d.log' % (Task._normalize(os.path.basename(test_binary)),
-                                 Task._normalize(test_name), execution_number)
+    # The log file name is a combination of a test binary name and a test name.
+    # If the log file name exceeds OS limitations it will be truncated.
+    MKSTEMP_RANDOM_LEN = 12  # length of mkstemp random string being added
```
pbos, line 213: "# Space reserved for mkstemp suffix."

gtest_parallel.py
```diff
@@ -204,14 +204,39 @@ def _logname(output_dir, test_binary, test_name, execution_number):
...
+      if not os.path.exists(log_name):
+        os.makedirs(output_dir, exist_ok=True)
+        with open(log_name, 'w+'):
+          pass
+        os.remove(log_name)
+    except OSError as os_error:
+      # truncate file name if error is Errno 36: File name too long
+      if os_error.errno == 36:
+        prefix_length = ((os.statvfs('.').f_namemax -
+                          MKSTEMP_RANDOM_LEN) if hasattr(os, 'statvfs') else MAX_PREFIX_LENGTH) - len(suffix)
```
pbos, line 231: I think this would read easier as: if hasattr(..): prefix_length = ..., i.e. avoid the ternary. as it's long.

gtest_parallel_tests.py
```diff
@@ -311,36 +311,44 @@ def test_times_worker():
...
+                                    'Instantiation/Test.case/2', 3))
+
+      self.assertEqual(
+        os.path.join(root, 'a', 'b', 'bin-Test_case-1.log'),
+        gtest_parallel.Task._logname(os.path.join(root, 'a', 'b'),
+                                    os.path.join(root, 'c', 'd', 'bin'),
+                                    'Test.case', 1))
+      MAX_PREFIX_LENGTH = 240
+      long_test_name = 'a' * (os.statvfs('.').f_namemax if hasattr(os,
+                                                                   'statvfs') else MAX_PREFIX_LENGTH + 1)
```
pbos, line 344: Could you make this one a separate test? Maybe `test_truncates_file_name`

Also I think this test only makes sense on systems that have `statvfs`, can it run conditionally? This test requires OSes without `statvfs` to reject filenames that are longer which is maybe not what we want. I'm OK with not testing the MAX_PREFIX_LENGTH branch.

pbos (commented):
Sorry for the slow response time, new parent here.

MARY203475 (approved):

PR #79, opened 2020-09-23
Add an option to print task status continuously

jay-zhuang:
Seems in non-tty mode the write is buffered. If the output data is small, it won't print out until all tests are done.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -300,7 +300,7 @@ class TaskManager(object):
 
 
 class FilterFormat(object):
-  def __init__(self, output_dir, print_task_status_continuously=False):
+  def __init__(self, output_dir):
     if sys.stdout.isatty():
       # stdout needs to be unbuffered since the output is interactive.
       if isinstance(sys.stdout, io.TextIOWrapper):
@@ -316,7 +316,6 @@ class FilterFormat(object):
 
     self.total_tasks = 0
     self.finished_tasks = 0
-    self.print_task_status_continuously = print_task_status_continuously
     self.out = Outputter(sys.stdout)
     self.stdout_lock = threading.Lock()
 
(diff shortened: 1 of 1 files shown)


pbos:
Can you see if this can detect whether it's interactive or not? `ls --color=auto` as an example won't print color codes if it's being piped somewhere else. I'd prefer that we did the same rather than add a flag.

jay-zhuang:
> Can you see if this can detect whether it's interactive or not? ls --color=auto as an example won't print color codes if it's being piped somewhere else. I'd prefer that we did the same rather than add a flag.

@pbos thanks for the quick response.

The problem is in CircleCI, the current implementation prints nothing while it's running, for example: https://app.circleci.com/pipelines/github/facebook/rocksdb/3536/workflows/a3559c67-7059-4a0a-9e6e-df1d6e4597c9/jobs/31722
image

Which is the same as if a test is hang and CircleCI may timeout after 10 minutes.

With this change, it's able to print the out line by line:
image

I guess CircleCI just pipe the std.output to web UI, but for the user, it's still is_a_tty() and does support color, as I can see the colored error message when it's failed:
image

Sorry, what do you mean this can detect whether it's interactive or not?

pbos:
I thought `isatty()` would cover it. Does piping `gtest-parallel` to `cat` solve the problem if you detect `isatty()`? I guess what I'm getting at is if there's a way to make it a non-`isatty()` for the CircleCI. For context, I find adding flags fairly expensive.

jay-zhuang:
> I thought `isatty()` would cover it. Does piping `gtest-parallel` to `cat` solve the problem if you detect `isatty()`? I guess what I'm getting at is if there's a way to make it a non-`isatty()` for the CircleCI. For context, I find adding flags fairly expensive.

Great, piping to `cat` works. The only problem now is the buffered output, it won't print out anything until it reaches the buffer size or all tests are done. Verified that adding a `flush()` would fix the problem. I updated the patch for that. Please take a look.

pbos:
Thank you! :)

PR #81, opened 2021-03-22
Add gtest_output support

tonka3000:
This PR add support for `gtest_output` xml format.

gtest support `--gtest_output` to save files as xml. An example would be `my.exe --gtest_output=xml:myfile.xml`. gtest-parallel does not support this option right now, but it would be very practical to have support for it because many CI systems like e.g. GitLab CI support artifact upload to visualize the test results in a project.

## How a user can use this option
I've added and `--gtest_output` option which is very similar to the original one.

Example: `--gtest_output xml:myfile.xml`

## How is it implemented
I save the gtest xml file for every exectuable run in temp files and join them together after all tasks have finished. This way I get the original content of the gtest xml output which be joined together.

Greetings
Michael

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -440,6 +440,8 @@ def join_gtest_output_files(files, output_filename):
   '''join given files into a single one'''
   from xml.dom import minidom
 
+  files = [f for f in files if os.path.exists(f) and os.path.getsize(f) > 0]
+
   base_file = files[0]
   del files[0]
   doc = minidom.parse(base_file)
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -779,7 +779,7 @@ def default_options_parser():
                     default=False, help='Do not run tests from the same test '
                                         'case in parallel.')
   parser.add_option('--gtest_output', type='string', default=None,
-                    help="Save the results to a file")
+                    help="Save the results to a file. Needs to start with 'xml:'")
   return parser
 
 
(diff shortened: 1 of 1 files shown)


tonka3000:
@googlebot I signed it!

JensHuthmann:
The merging of the xml fails, if one of the tests crashes the test itself. Which happens for me with glog and CHECK(...). This creates a SIGABRT. The merge fails because no xml is created for that test.

philipogorman:
this would be a great feature for running tests on jenkins.

tonka3000:
@JensHuthmann @pbos 

I've merged the master branch and resolved the conflicts.

I also skip now files which are empty (0 bytes) when e.g. `abort()` is called in a test. This is not a great solution because the existence of the test is now fully hidden from the xml file, but I'm not quite sure how to handle it better. The gtest version I use (1.8.x) don't write a xml file when one of the tests send SIGABRT. Not quite sure if more recent version of gtest make a better job here.

Suggestions are welcome.

Greetings
Michael

gtest_parallel.py
```diff
@@ -436,6 +436,57 @@ def add_stats(stats, task, idx):
   def flush(self):
     self.out.flush_transient_output()
 
+def join_gtest_output_files(files, output_filename):
+  '''join given files into a single one'''
+  from xml.dom import minidom
+
+  files = [f for f in files if os.path.exists(f) and os.path.getsize(f) > 0]
```
pbos, line 443: Can you insert a TODO to revisit how empty files / crashes are handled and highlight that we skip that for now?

Also could we make sure to print a summary of how many tests failed to generate output? I think this may go through list of files before stripping out the !exists/empty ones. It would be OK to print every single failure here I think and not necessarily print x/y tests aborted. Can we get the test name corresponding to the failure as well?

pbos:
Sorry for the slow response I've been more busy than I'd like. Can we make sure this error handling is explicit (by comments and printing which tests aborted and that they're therefore not present in the xml output)?

tonka3000:
Hi @pbos ,

I will check it (hopefully in the next couple of days).

mr1sunshine:
Such a helpful feature!

@tonka3000 I'm just wondering if you had a chance or if you are planning to finish a work on this PR, so it gets merged?

tonka3000:
Hey @mr1sunshine ,

never found the time to finish the PR. Work drifted to different technology and I used gtest rarely since the shift.
When somebody want to finish this in the meantime, feel free.

PR #85, opened 2022-04-14
Express json output tags time and times as seconds.

jleconte2:
The json specification says that 'time' and 'times' should be expressed as seconds.
https://chromium.googlesource.com/chromium/src/+/HEAD/docs/testing/json_test_results_format.md

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -270,7 +270,7 @@ class TaskManager(object):
     self.times.record_test_time(task.test_binary, task.test_name,
                                 task.last_execution_time)
     if self.test_results:
-      self.test_results.log(task.test_name, int(task.runtime_ms / 1000),
+      self.test_results.log(task.test_name, task.runtime_ms / 1000,
                             "PASS" if task.exit_code == 0 else "FAIL")
 
     with self.lock:
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -270,7 +270,7 @@ class TaskManager(object):
     self.times.record_test_time(task.test_binary, task.test_name,
                                 task.last_execution_time)
     if self.test_results:
-      self.test_results.log(task.test_name, task.runtime_ms,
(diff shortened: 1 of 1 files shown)


gtest_parallel.py
```diff
@@ -270,7 +270,7 @@ def __register_exit(self, task):
     self.times.record_test_time(task.test_binary, task.test_name,
                                 task.last_execution_time)
     if self.test_results:
-      self.test_results.log(task.test_name, task.runtime_ms,
+      self.test_results.log(task.test_name, int(task.runtime_ms / 1000),
```
pbos: This truncation seems undesirable (for very fast tests 0/1 seconds is not very helpful), can this print 0.543 seconds for instance?

pbos (changes requested):
I remember running this with sub-second tests and still having the runtime be useful, json numbers are decimals so we should be able to log 0.5 seconds, correct?

jleconte2:
Done, I think decimal should work fine.

gtest_parallel.py
jleconte2: Done, I think decimal should work fine.

gtest_parallel.py
```diff
@@ -270,7 +270,7 @@ def __register_exit(self, task):
     self.times.record_test_time(task.test_binary, task.test_name,
                                 task.last_execution_time)
     if self.test_results:
-      self.test_results.log(task.test_name, task.runtime_ms,
+      self.test_results.log(task.test_name, task.runtime_ms / 1000,
```
pbos: There's some code in here that assumes that this script works with Python2 and Python3 both, I'm not sure if that's true anymore, but could you make this / 1000.0 to avoid truncation just in case Python2 happens to still work? Other than that this is good to merge. Thank you!

jleconte2: Good point, I already switched to full python3 mode.

pbos:
Thanks, please send me a screenshot of reasonable test times if it works! :)

PR #87, opened 2022-05-27
Consider tests that reach the timeout as failed.

jleconte2:


gtest_parallel.py
```diff
@@ -446,11 +451,18 @@ def __init__(self, json_dump_filepath):
...
             "PASS": 0,
             "FAIL": 0,
+            "TIMEOUT": 0,
         },
         "tests": {},
     }
 
-  def log(self, test, runtime_seconds, actual_result):
+  def log(self, test, runtime_seconds, exit_code):
+    if exit_code is None:
```
pbos, line 460: I like this. No exit code resolving to timeout seems pretty alright.

PR #90, opened 2022-06-20
Make gtest-parallel to use python3

bc-lee:
According to the Google Python style guide [1] and the Chromium Python
style guide [2], the shebang should be #!/usr/bin/env python3.
This is useful for running WebRTC unittests in python2-less environment.

[1] https://google.github.io/styleguide/pyguide.html
[2] https://chromium.googlesource.com/chromium/src/+/refs/heads/main/styleguide/python/python.md


pbos:
Thank you, good catch!

PR #94, opened 2023-07-31
Add per test timeout

Momcilo42:
This allows for a test to timeout when hanging.
It's also more useful than the general timeout option as this shows if more than one test can timeout.

### DIFF
--- a/gtest_parallel.py
+++ b/gtest_parallel.py
@@ -84,12 +84,17 @@ class SigintHandler(object):
     with self.__lock:
       return self.__got_sigint
 
-  def wait(self, p):
+  def wait(self, p, timeout_per_test):
     with self.__lock:
       if self.__got_sigint:
         p.terminate()
       self.__processes.add(p)
-    code = p.wait()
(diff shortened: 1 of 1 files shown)


Momcilo42:
@pbos Can you take a look?

pbos:
Hi, sorry I'm out on parental leave until ~November. I'll take a look then.

@jleconte2 do you know off the top of your head how WebRTC deals with test timeouts? I wouldn't want us to duplicate functionality if we already offer something similar.

jleconte2:
WebRTC runs gtest_parallel under LUCI infrastructure which provides its own timeouts management.
It currently does not use the timeouts provided by gtest_parallel.

Momcilo42:
@pbos Can you take a look now?

pbos:
It looks like upstream isn't interested in adding per-test-timeout (since gtest runners like blaze and cmake support timeout) and I think gtest-parallel falls within the scope of being a gtest runner. Thanks!

